PR #8, opened 2016-11-16
Feature/mina

d4be4st:
for #1

### DIFF
--- a/mina_deploy.rb
+++ b/mina_deploy.rb
@@ -3,6 +3,7 @@ require 'mina/git'
 require 'mina/secrets'
 require 'mina/infinum'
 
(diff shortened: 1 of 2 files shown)


mina_deploy.rb
```diff
@@ -0,0 +1,39 @@
+require 'mina/rails'
+require 'mina/git'
+require 'mina/secrets'
+require 'mina/infinum'
+
+set :application_name, 'awesome_app'
+set :repository, 'git://...'
+set :user, 'deploy'
+
+def staging
```
DamirSvrtan: Shouldn't this be `task :staging do`?

d4be4st: good catch :D

PR #12, opened 2016-11-18
readme file

vr4b4c:


vr4b4c:
resolves #4

template.rb
```diff
@@ -1,6 +1,52 @@
 require 'net/http'
 
-create_file 'README.md', 'Development: run ./bin/setup', force: true
+README_MD = <<-HEREDOC.strip_heredoc
+# README
+
+## Dependent projects
+  * [Single sign-on service](https://github.com/infinum/accounts)
```
DamirSvrtan: So this way, devs should erase some stuff from the Readme once they run this template. I'd maybe add it but commented out. And probably add https://www.vaultproject.io/ uncommented since we are using it on 99% of stuff.

template.rb
```diff
@@ -1,6 +1,52 @@
...
+
+## Binaries
+### System
+  * node, eslint
+
+### Gem
+  * vault
```
DamirSvrtan: I'd add Vault in the dependent projects, not in gems. I don't think we need the Gem section since you can overview it in the Gemfile.

template.rb
```diff
@@ -1,6 +1,52 @@
 require 'net/http'
 
-create_file 'README.md', 'Development: run ./bin/setup', force: true
+README_MD = <<-HEREDOC.strip_heredoc
+# README
+
+## Dependent projects
```
DamirSvrtan: I'd call this `Dependencies` and add 2 sections:

## Dependencies
#### Projects
#### System

template.rb
```diff
@@ -1,6 +1,52 @@
...
+```bash
+rspec
+```
+
+## Deployment
+Environments:
+  * staging <stg>: [staging](https://staging.com)
+  * production <prod>: [production](https://production.com)
+
+Run:
```
DamirSvrtan: I'd add comments here how everything should be deployed through SemaphoreCI, and in rare severe cases be deployed through mina with the mentioned command.

template.rb
```diff
@@ -1,6 +1,52 @@
...
+```
+mina <env> deploy
+```
+HEREDOC
+
+create_file 'README.md', README_MD
+remove_file 'README.rdoc'
```
DamirSvrtan: Rails 5 now adds a Markdown file, you need to add `force: true` as a third argument to the `create_file` method and don't need the remove_file call.

template.rb
vr4b4c: [Dependent Projects] and [Dependencies -> System] are merged under [Dependencies]([Dependencies -> Gems] are removed)

template.rb
vr4b4c: @DamirSvrtan edit section to your desire, nonetheless environments should be listed in any case

template.rb
vr4b4c: done

PR #21, opened 2017-11-21
Feature/rspec

d4be4st:
Add rspec-rails gem and installs rspec files


template.rb
```diff
@@ -262,6 +263,8 @@
 
 run 'bundle exec secrets init'
 
+run 'budnle exec rails generate rspec:install'
```
vr4b4c: `budnle` -> `bundle`

PR #23, opened 2017-11-21
Feature/interactive

d4be4st:
I have added an interactive component to template

Copied from https://github.com/mattbrictson/rails-template/blob/master/template.rb :D

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -264,7 +264,7 @@ run 'bundle install'
 
 run 'bundle exec secrets init'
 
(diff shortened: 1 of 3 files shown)


template.rb
```diff
@@ -262,6 +264,17 @@
 
 run 'bundle exec secrets init'
 
+run 'budnle exec rails generate rspec:install'
```
melcha: Typo: budnle -> bundle

PR #24, opened 2017-11-21
create ruby-version through rbenv

d4be4st:
Setting .ruby-version

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -264,10 +264,10 @@ append_file '.gitignore', GITIGNORED_FILES
 
 say('Updating ruby-build and installing latest ruby', :cyan)
 
-if !run('brew info ruby-build', capture: true).include?('Not installed')
-  run 'brew upgrade ruby-build'
-else
+if run('brew info ruby-build', capture: true).include?('Not installed')
   run 'cd "$(rbenv root)"/plugins/ruby-build && git pull'
+else
+  run 'brew upgrade ruby-build'
 end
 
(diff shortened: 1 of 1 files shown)


melcha:
I agree with @DamirSvrtan, I think we should somehow enforce the newest Ruby version. 

I suggest asking a question at the top of the template, something like this:
```
if yes?(“Rbenv global returned 2.4.2. Is this the latest stable Ruby version?")
  # do everything
else
    puts “Please install the latest Ruby version and set it as global."
end
```

It would be great if we could fetch the latest Ruby version via some API and compare it to the rbenv global output, but this will also work.

d4be4st:
Ok, new take @infinum/backend-team !

Template now tries to update ruby-build. As i know half of us have it installed through homebrew and other half as a plugin i have taken precautions.

After ruby-build update, rbenv finds latest stable ruby build and installs it.

template.rb
```diff
@@ -259,22 +259,44 @@
 append_file '.gitignore', GITIGNORED_FILES
 
 # Finish
+
+## Install latest ruby and sets it as local version
+
+say('Updating ruby-build and installing latest ruby', :cyan)
+
+if !run('brew info ruby-build', capture: true).include?('Not installed')
```
DamirSvrtan: Everything is fine I'd just reverse the order of if else so you don't have to use negations in the first clause.

DamirSvrtan:
@infinum/backend-team if there aren't any comments I'd advise @d4be4st to merge it in.

PR #26, opened 2017-11-27
Add database yml config, resolves #20

### LINKED ISSUE #20: Update the keys in database.yml to reflect secrets.yml

d4be4st:


template.rb
```diff
@@ -196,12 +218,17 @@
...
 
 FIGARO_FILE = <<-HEREDOC.strip_heredoc
+  database_host: localhost
+  database_username: postgres
+  database_password: ""
   bugsnag_api_key: ADD_IT_HERE
 
   development:
     secret_key_base: #{SecureRandom.hex(64)}
+    database_database: #{app_name}_development
```
DamirSvrtan: wouldn't it make sense to call this key `database_name`?

d4be4st: I use `database` because that is used in `database.yml`, not `name`

vr4b4c: ++ for `database_name`

DamirSvrtan: Let's change this to databae_name and merge it in.

template.rb
```diff
@@ -177,6 +194,11 @@
 SECRETS_YML_FILE = <<-HEREDOC.strip_heredoc
   default: &default
     secret_key_base: <%= Figaro.env.secret_key_base! %>
+    database:
+      database: <%= Figaro.env.database_name! %>
```
DamirSvrtan: You have to change it here also @d4be4st

PR #28, opened 2017-12-07
Separate application.yml

d4be4st:
Also added default_url host, needed for letter_opener


d4be4st:
@melcha skuzila, diz does not work :(

PR #32, opened 2017-12-21
Fixes #31

### LINKED ISSUE #31: When installing new ruby, rbenv local does not switch in current shell

d4be4st:
This is one solution

Another would be to check for ruby version at the very beginning of the script and exit if current ruby version is not the latest.

This will present a problem even if you have the latest ruby version installed but is not set as global.

Thoughts @infinum/backend-team

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -304,13 +304,13 @@ run "rbenv install -s #{latest_ruby}"
 run "rbenv local #{latest_ruby}"
 
 ## Bundle install
-run 'bundle install'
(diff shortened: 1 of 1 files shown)


DamirSvrtan:
Interesting solution, however I'm leaning to a simpler solution...just write down in the README 'install the latest stable version of Ruby, then run the rails generator...' and that's it. 👍 for agreeing on this, 👎 for disagreeing.

PR #42, opened 2019-11-12
Update rubocop cop rules

JureCindro:
#### Aim
Update rubocop cop rules.

#### Solution
- added rubocop-rails gem and require rubocop-rails cops (https://github.com/rubocop-hq/rubocop/issues/5976)
- added departments for all cops (https://github.com/rubocop-hq/rubocop/pull/7385)
- changed word and symbol arrays cops from disabled to enforcing style brackets, forcing us to using a consistent style across the app instead of letting us use both. 
- removed disabling of mutable constants cop, constants should be constant 🤓

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,32 +1,31 @@
-require: rubocop-rspec
+require:
+  - rubocop-rspec
+  - rubocop-rails
 
-LineLength:
+Metrics/LineLength:
   Max: 100
 
-Documentation:
+Metrics/BlockLength:
+  Exclude:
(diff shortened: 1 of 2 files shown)


.rubocop.yml
```diff
@@ -1,33 +1,32 @@
...
+Style/SymbolArray:
+  EnforcedStyle: brackets
 
 Rails:
   Enabled: true
 
 RSpec/ExampleLength:
   Max: 10
 
-Style/MutableConstant:
```
d4be4st: dont really know why this was here :D

PR #43, opened 2019-11-15
Add brakeman to overcommit

d4be4st:
Add brakeman to overcommit

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -176,7 +176,7 @@ append_to_file 'Gemfile', after: "group :development do\n" do
   gem 'mina-infinum', require: false
   gem 'overcommit', require: false
   gem 'rubocop', require: false
-  gem 'rubocop-rspec', require: false  
+  gem 'rubocop-rspec', require: false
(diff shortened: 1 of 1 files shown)


d4be4st:
@infinum/backend-team adding brakeman to overcommit

template.rb
```diff
@@ -275,6 +275,10 @@
 
   HardTabs:
     enabled: true
+
+PrePush:
+  Brakeman:
+    enabled: true
```
uncoverd: Me and @JureCindro were looking into this yesterday, this works if you only have 1 version of breakman. The better solution would be to add Breakman to the Gemfile and add the command parameter (this also applies to Rubocop)
```
  RuboCop:
    enabled: true
    on_warn: fail
    command: ['bundle', 'exec', 'rubocop']
```

This would also allow developers to use breakman in development (or even add it as a CI build step). What do you think?

d4be4st: Interesting, i did not think about this..
I must've assumed that overcommit already does this

Added bundle exec to all three

PR #44, opened 2019-11-28
Use secrets_cli from gemfile

melcha:
#### Aim
Use secrets_cli from the Gemfile for all secrets handling during build and deployment, so we don't need to manually install secrets_cli gem on the server for each ruby version, and so we all use same secrets_cli version per project.

#### Solution
I updated the deploy script and moved secrets_cli outside of the development group in the Gemfile.

### DIFF
--- a/mina_deploy.rb
+++ b/mina_deploy.rb
@@ -32,7 +32,7 @@ task :deploy do
     invoke :'git:clone'
     invoke :'deploy:link_shared_paths'
     invoke :'bundle:install'
-    command "bundle exec secrets pull -e #{fetch(:secrets_env)} -y"
+    invoke :'secrets:pull'
     invoke :'rails:db_migrate'
     invoke :'rails:assets_precompile'
     invoke :'deploy:cleanup'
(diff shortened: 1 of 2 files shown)


borislavgoles:
This makes total sense and benefits all parties included

mina_deploy.rb
```diff
@@ -21,17 +22,17 @@
...
 end
 
 task :deploy do
   invoke :'git:ensure_pushed'
   deploy do
     invoke :'git:clone'
-    invoke :'secrets:pull'
     invoke :'deploy:link_shared_paths'
     invoke :'bundle:install'
+    command "bundle exec secrets pull -e #{fetch(:secrets_env)} -y"
```
d4be4st: i would rather update https://github.com/infinum/mina-secrets/blob/master/lib/mina/secrets/tasks.rb
to use bundle exec. Also we can update mina-secret to use secret_env || rails_env

melcha: Done. I'll revert this to invoke after we publish new mina-secrets version.

d4be4st: you can revert this and move secrets:pull down below bundle:install

PR #47, opened 2020-03-09
Add rubocop binstubs and move to development group

JureCindro:
Task: NA

#### Aim
Spring binstubs are used only in the development enviroment so the additional binstub gems should be moved to the development group only.

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -353,9 +353,10 @@ run 'bundle exec rails generate rspec:install'
 
 ## Initialize spring
 if yes?('Install spring? [No]', :green)
-  append_to_file 'Gemfile', after: "group :development, :test do\n" do
+  append_to_file 'Gemfile', after: "group :development do\n" do
     <<-HEREDOC
     gem 'spring-commands-rspec'
+    gem 'spring-commands-rubocop'
     HEREDOC
   end
   run 'bundle install'


template.rb
```diff
@@ -353,9 +353,10 @@
 
 ## Initialize spring
 if yes?('Install spring? [No]', :green)
-  append_to_file 'Gemfile', after: "group :development, :test do\n" do
+  append_to_file 'Gemfile', after: "group :development do\n" do
     <<-HEREDOC
     gem 'spring-commands-rspec'
+    gem 'spring-commands-rubocop'
```
d4be4st: Interesting but this wont be used by overcommit right? https://github.com/infinum/default_rails_template/blob/master/template.rb#L269

JureCindro: Yeah, we could change the command to:
```yml
    command: ['bin/rubocop']
```
But I am not sure about the performance improvements. Looking at `rubocop-rails` it does not seem to load the rails app, so the only improvements would probably be in the differences in bundle loading, but at the cost of having a slower first run of rubocop, if spring was not running before, as it loads the whole rails app which is actually not required for rubocop. 

TL;DR:
I'm guessing binstubbing rubocop has a net negative effect on run time on projects with smaller bundles, and would probably make sense not to include it in the template. I would like to test this theory on a project with an enormous bundle, to see if it holds water.

PR #49, opened 2020-05-21
Add zeitwerk check to overcommit pre-push

melcha:
With custom readable error messages. Thanks @JureCindro !

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -283,6 +283,11 @@ PrePush:
   Brakeman:
     enabled: true
     command: ['bundle', 'exec', 'brakeman']
+
+  ZeitwerkCheck:
+    enabled: true
+    description: 'Checks project structure for Zeitwerk compatibility'
+    command: ['bundle', 'exec', 'rails zeitwerk:check']
 HEREDOC
 create_file '.overcommit.yml', OVERCOMMIT_YML_FILE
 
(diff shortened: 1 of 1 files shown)


d4be4st:
we also can add `run 'overcommit --sign pre-push'` to the bottom of the template

template.rb
```diff
@@ -324,6 +329,31 @@
...
 
+# .git-hooks/pre_push/zeitwerk_check.rb
+ZEITWERK_CHECK_FILE = <<-HEREDOC.strip_heredoc
+# frozen_string_literal: true
+
+module Overcommit
+  module Hook
+    module PrePush
+      class ZeitwerkCheck < Base
+        # rubocop:disable Style/StringLiterals
```
JureCindro, line 340:
```suggestion
```
This is not needed, right?

melcha: It is, I ran rails new with this change to see if it works, and rubocop was complaining about "".

JureCindro: Ah I see, it's because it actually inserts a new line in the ruby file.
```ruby
def run
  result = execute(command)
  return :pass if result.success?

  extract_messages result.stderr.split("
"),
                   /^expected file (?<file>[[:alnum:]].*.rb)/
end
```

This patch should do the trick I belive:
```diff
diff --git a/template.rb b/template.rb
index 3490fba..2663b7f 100644
--- a/template.rb
+++ b/template.rb
@@ -337,15 +337,13 @@ module Overcommit
   module Hook
     module PrePush
       class ZeitwerkCheck < Base
-        # rubocop:disable Style/StringLiterals
         def run
           result = execute(command)
           return :pass if result.success?
 
-          extract_messages result.stderr.split("\n"),
+          extract_messages result.stderr.split("\\n"),
                            /^expected file (?<file>[[:alnum:]].*\.rb)/
         end
-        # rubocop:enable Style/StringLiterals
       end
     end
   end
```

melcha: Yep, this works! Thanks

PR #53, opened 2020-07-22
Update system dependencies chapter

lovro-bikic:
Task: [#217](https://app.productive.io/1-infinum/tasks/1116586)

#### Aim

To have all system dependencies necessary to run the application defined along with required versions.

#### Solution

Updated system dependencies chapter. Replaced npm commands with Yarn commands (it's the default package manager for Rails, see more [here](https://guides.rubyonrails.org/5_1_release_notes.html#yarn-support)).

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -89,8 +89,8 @@ BIN_SETUP = <<-HEREDOC.strip_heredoc
     system 'gem install bundler --conservative'
     system 'bundle check || bundle install'
 
-    # puts '== Installing node modules =='
-    # system 'npm install'
+    # puts '== Installing JS dependencies =='
+    # system 'yarn install'
 
     puts "== Installing overcommit =="
     system 'overcommit --install'
@@ -119,11 +119,8 @@ BIN_UPDATE = <<-HEREDOC.strip_heredoc
     system 'gem install bundler --conservative'
     system 'bundle check || bundle install'
 
-    # puts '== Installing node modules =='
-    # system 'npm install'
-
-    # puts '== Building frontend =='
-    # system 'npm run build'
+    # puts '== Installing JS dependencies =='
+    # system 'yarn install'
 
     puts '== Pulling secrets =='
     system 'bundle exec secrets pull'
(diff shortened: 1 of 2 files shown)


template.rb
```diff
@@ -11,7 +11,9 @@
 ## Dependencies
 
 ### System
-  * yarn, eslint
+  * Ruby (defined in .rbenv file)
```
d4be4st, line 14: .ruby-version ?

lovro-bikic: Yep. My mistake.

d4be4st:
rake assets:precompile does yarn install

melcha:
Yes, but it doesn't exit on error if something can't be installed - this happened on Loop a few weeks ago. And if we add it right after bundle install, I'm guessing we won't get those `mjml is not installed, npm install mjml` warning/error messages.

d4be4st:
mjml might be your only reason why you would need to do yarn install before :)

because if you do manually yarn install then it does it twice, once manually once in rake assets precompile

uncoverd:
I wouldn't put the weight of knowing whether a NPM package needs a binary to function on the developer's mind. I think most of the time we're not that intimately aware of what JS dependencies are actually doing behind the scenes. If we can make it human-proof then we should. The second time yarn install is run will be a no-op and take a few seconds tops, so I would vote we add the `yarn install` step.

melcha:
I agree with @uncoverd. 
I'm currently deploying Cormeum for the first time from Semaphore, and the deploy failed without a clear error message:

```bash
-----> Precompiling asset files

       RAILS_ENV=staging environment is not defined in config/webpacker.yml, falling back to production environment

 !     ERROR: Deploy failed.

-----> Cleaning up build

       Unlinking current

       OK

       Connection to cormeum.byinfinum.co closed.
```

After adding `command 'yarn install'` to deploy.rb, the error was: ```bash: line 99: yarn: command not found```

nikajukic:
I agree with @uncoverd and @melcha. `yarn install` would have saved us a bunch of time while deploying Loop to new architecure.

PR #63, opened 2021-05-13
[#350] Add deploy and build scripts

d4be4st:
Task: [#350](https://app.productive.io/1-infinum/tasks/1642611)

#### Aim

We need to standardize our build and deploy scripts

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -135,6 +135,114 @@ BIN_UPDATE = <<-HEREDOC.strip_heredoc
 HEREDOC
 create_file 'bin/update', BIN_UPDATE, force: true
 
+BIN_BUILD = <<~HEREDOC.strip_heredoc
+  #!/usr/bin/env bash
+
+  set -o errexit
+  set -o pipefail
+  set -o nounset
+
+  echo "=========== setting env variables ==========="
+  export RAILS_ENV='test'
+  export BUNDLE_APP_CONFIG='.bundle/ci-build'
+
+  echo "=========== install bundler ==========="
+  bundler_version=`tail -n 1 Gemfile.lock | xargs`
+  time gem install bundler:$bundler_version --no-document
+
+  echo "=========== bundle install ==========="
+  time bundle install
+
+  echo "=========== bundle audit ==========="
+  time bundle exec bundle-audit check --update
+
+  echo "=========== zeitwerk check ==========="
+  time bundle exec rails zeitwerk:check
+
+  echo "=========== brakeman ==========="
+  time bundle exec brakeman
+
+  echo "=========== rubocop  ==========="
+  time bundle exec rubocop --format simple
+
+  echo "=========== tests ============="
+  echo "=========== secrets pull ============="
+  time bundle exec secrets pull -e development -y
+
+  echo "=========== rails db:test:prepare ==========="
+  time bundle exec rails db:test:prepare
+
+  echo "=========== rspec ==========="
+  time bundle exec rspec
+HEREDOC
+create_file 'bin/build', BIN_BUILD, force: true
+
+BIN_DEPLOY = <<~HEREDOC.strip_heredoc
+  #!/usr/bin/env bash
+
+  set -o errexit
+  set -o pipefail
+  set -o nounset
+
+  echo "=========== setting env variables ==========="
+  environment=$1
+  export RAILS_ENV='test'
+  export BUNDLE_APP_CONFIG='.bundle/ci-deploy'
+
+  echo "=========== install bundler ==========="
+  bundler_version=`tail -n 1 Gemfile.lock | xargs`
+  time gem install bundler:$bundler_version --no-document
+
+  echo "=========== bundle install ==========="
+  time bundle install
+
+  echo "=========== mina deploy =============="
+  time bundle exec mina $environment ssh_keyscan_domain deploy
+
+  #############################################
+  # Uncomment this if you need to publish dox #
+  # Delete not needed environment             #
+  #############################################
+  # if [[ $environment =~ ^(production|uat|staging)$ ]]; then
+  #   echo "=========== secrets pull =============="
+  #   time bundle exec secrets pull -e development -y
+
+  #   echo "=========== yarn install =============="
+  #   time yarn install
+
+  #   echo "=========== rails db:test:prepare ==========="
+  #   time bundle exec rails db:test:prepare
+
+  #   echo "=========== mina dox publish ==========="
+  #   time bundle exec mina $environment dox:publish
+  # fi
+HEREDOC
+create_file 'bin/deploy', BIN_DEPLOY, force: true
+
+# bundler config
+BUNDLER_CI_BUILD_CONFIG = <<~HEREDOC.strip_heredoc
+  ---
+  BUNDLE_DEPLOYMENT: "true"
+  BUNDLE_WITHOUT: "development deploy"
+HEREDOC
+create_file '.bundle/ci-build/config', BUNDLER_CI_BUILD_CONFIG, force: true
+
+BUNDLER_CI_DEPLOY_CONFIG = <<~HEREDOC.strip_heredoc
+  ---
+  BUNDLE_DEPLOYMENT: "true"
+  BUNDLE_WITHOUT: "default development test ci"
+  # use line below when using dox
+  # BUNDLE_WITHOUT: "development"
+HEREDOC
+create_file '.bundle/ci-deploy/config', BUNDLER_CI_DEPLOY_CONFIG, force: true
+
+BUNDLER_SERVER_CONFIG = <<~HEREDOC.strip_heredoc
+  ---
+  BUNDLE_DEPLOYMENT: "true"
+  BUNDLE_WITHOUT: "development test ci deploy"
+HEREDOC
+create_file '.bundle/server/config', BUNDLER_SERVER_CONFIG, force: true
+
 # bugsnag
 BUGSNAG_CONFIG = <<-HEREDOC.strip_heredoc
   Bugsnag.configure do |config|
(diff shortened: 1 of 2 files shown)


d4be4st:
@nikajukic when you open a task can you please remind me to update the description of this PR :D

template.rb
```diff
@@ -135,6 +135,111 @@
...
+  time bundle exec bundle-audit check --update
+
+  echo "=========== zeitwerk check ==========="
+  time bundle exec rails zeitwerk:check
+
+  echo "=========== brakeman ==========="
+  time bundle exec brakeman
+
+  echo "=========== rubocop  ==========="
+  time bundle exec rubocop .
```
JureCindro, line 166: Why is the dot needed?

```suggestion
  time bundle exec rubocop --format simple
```
This does not output dots but does describe the errors.

template.rb
```diff
@@ -135,6 +135,111 @@
...
+
+  echo "=========== tests  ============="
```
JureCindro, line 168:
```suggestion
  echo "=========== tests ============="
```

template.rb
```diff
@@ -135,6 +135,111 @@
...
+
+  echo "=========== install bundler ==========="
+  bundler_version=`tail -n 1 Gemfile.lock | xargs`
+  time gem install bundler:$bundler_version --no-document
+
+  echo "=========== bundle install ==========="
+  time bundle install
+
+  echo "=========== bundle audit ==========="
+  time bundle exec bundle-audit check --update
```
JureCindro, line 157: So in case of a CVE you would be required to:

1. create a PR where you ignore the CVE and immediately merge into master, so deploys are not blocked.
2. rebase all ongoing branches on updated master
3. merge-squash to staging/uat to test the updated gem 
4. once the dependency update is greenlighted you can remove the ignore CVE directive

This seems quite disruptive. Once a CVE is found, everything will stop working.

template.rb
```diff
@@ -135,6 +135,111 @@
...
+
+  echo "=========== tests  ============="
+  echo "=========== secrets pull ============="
+  time bundle exec secrets pull -e development -y
+
+  echo "=========== rails db:test:prepare ==========="
+  time bundle exec rails db:test:prepare
+
+  echo "=========== rspec ==========="
+  time bundle exec rspec
```
JureCindro, line 176:
```suggestion
  time bundle exec rspec --format failures
```
Does not output dots but it does print failures. It is a bit faster than the default dot progress (seemed like 10% faster locally for me)

template.rb
```diff
@@ -135,6 +135,111 @@
...
+  # time bundle exec secrets pull -e development -y
+
+  # echo "=========== yarn install =============="
+  # time yarn install
+
+  # echo "=========== rails db:test:prepare ==========="
+  # time bundle exec rails db:test:prepare
+
+  # echo "=========== mina dox publish ==========="
+  # time bundle exec mina $environment dox:publish
```
JureCindro, line 215: We don't usually need documentation on production deploys, right?
```
if [[ ${ENVIRONMENT} != 'production' ]]; then
  echo "=========== mina dox publish ==========="
  time bundle exec secrets pull -e development -y
  time bundle exec rails db:test:prepare
  time yarn global add aglio
  time bundle exec mina "${ENVIRONMENT}" dox:publish
fi
  ```

template.rb
uncoverd: That's a feature, not a bug 😂 How does not failing the build differ? You'll still have to open a CVE branch, test the fix on staging, deploy to production and rebase all pre-existing branches. The only difference seems to be that you have to decide when you deploy "I'll deal with this in this PR" or "I'll ignore it and come back to it". 

We like to think that these security updates aren't really important or urgent, but you only need 1 script kiddy hunting the web for vulnerable apps with a known CVE to achieve RCE and get you on the front pages and in legal trouble with a client. Imagine if NLBL, Atlas or Loop got breached and DBs deleted because it took us 5 days to update Rails for a security vulnerability. It's an uncomfortable thought but these things do happen. I think whatever collective pain we may face will pay dividends in assurance that we're actually doing the best we can. I wouldn't want to pay tens of thousands of dollars for an app and have someone miss a security update because it wasn't convenient to break a build 😢

d4be4st: We had the same conversation in https://github.com/infinum/rails-jsonapi-example-app/pull/42 :D

template.rb
d4be4st: This i would guess be different for different projects. We on underline have a `== 'staging'` as we have UAT but docs only on staging.

I do not have a preference here :D

uncoverd: How about we make it easy for the next poor soul and just provide an array to show the syntax? There are usecases where we may want to expose documentation on production. To be honest I dont know why we don't generate it for UAT?

if env in [staging, uat, production]
  do stuff
fi

template.rb
JureCindro: > We like to think that these security updates aren't really important or urgent, but you only need 1 script kiddy hunting the web for vulnerable apps with a known CVE to achieve RCE and get you on the front pages and in legal trouble with a client. 

I'm not saying that at all. I'm just saying that maybe we should rely on our due diligence rather than having to deal with disruptions to the work flow. Github has the security vulnerability scanner as well, why aren't we using that as well?

>Imagine if NLBL, Atlas or Loop got breached and DBs deleted because it took us 5 days to update Rails for a security vulnerability.

So how does this help prevent that? Because you do ignore the CVE in step 1 so that other work can continue while the fix is tested on staging. Nothing prevents you to forget about it. Or should we just stop working until the CVEs are all fixed and tested?

Don't get me wrong, I agree that CVEs need to be fixed asap. But this check has a quite easy but a bit annoying workaround, and in the end, I don't think there will be much difference. 
Because we have seen that when things go red, a lot of times the solution is to disable the red check (`overcommit --uninstall`). Well now, we're essentially just swapping this for `ignore: CVE-xxx` in a file.

I don't like crime, but I strongly disagree with putting petty thieves in jail for 10 years. Maybe the problem is not that we do not police this enough, but that people don't take the time to fix this, or can't take the time because of tight estimations, high work pressure, too high delivery expectations.

uncoverd: > I'm just saying that maybe we should rely on our due diligence rather than having to deal with disruptions to the work flow

We wouldn't be where we are now if we were doing our due diligence team wide. We have dozens of projects with open security vulnerabilities, production ones as well and its not because we did a good job raising the risks to the stake holders and making sure these get addressed.

>So how does this help prevent that? Because you do ignore the CVE in step 1 so that other work can continue while the fix is tested on staging. Nothing prevents you to forget about it.

Well hopefully it was enough of a disruption that you figured out you need to do something about it, someone else approved the ignore and you both took care of opening a task and assigning it a priority

> Or should we just stop working until the CVEs are all fixed and tested?

Probably yes. If we say that we provide a professional service and treat security as one of our core values we should make this one of the top priorities.

> But this check has a quite easy but a bit annoying workaround, and in the end, I don't think there will be much difference.

I'm not saying we need to keep it forever and thats its a complete fix. It's a small experiment that may or may not raise our development practice standards. If you have any other suggestions how to fix the culture apart from the other 4 ones I mentioned on Slack then feel free to voice them out. 

> I belive there are 4 ways to influence the individual developers to take the high road and we’re kinda tackling all of them
>give them a strong incentive (CI build fails) and teammate PR reviews with higher standards
>educatating them (workshop incoming) and their PMs and make it easy to do the right thing
>give them additional tooling (e.g. revisor daily checks) to make the problem more visible
> Share success stories and lessons learned in the team. We share a lot of the dependencies and yet we never talk about what to look out for, when a new Rails version is out, etc. For example the latest Rails action pack CVE - it would have been trivial to update Rails for others if the first person to do it documented in plain speak what to look out for, show the application code diffs, their deployment strategy, etc.

>Maybe the problem is not that we do not police this enough, but that people don't take the time to fix this, or can't take the time because of tight estimations, high work pressure, too high delivery expectations.

I'd say this is more internal project team policing and keeping each other honest. Up until now this was "policed" only if you had a project team member who cared deeply about this (there aren't that many to go around).

If the last three things you mentioned are true than fixing those should be a priority on that specific project. If the development team can't spend an hour or two to verify a critical security fix on the same day that it's announced we're doing something really wrong. Maybe this is the catalyst you need 😄

template.rb
JureCindro: Ugh, something I just realised. This format does not echo the seed so there is a flaky failure it will be next to impossible to debug. I guess we need the dots formatter.

d4be4st: How often do we need seed? 

Maybe we can find an external formatter that servers our purpose? or write our own? :)

uncoverd: We need need it on Loop for example because theres parallel workers there and the exact seed + order matters. I don't think I've ever needed the exact CI seed on a non-Loop project.

I'm curious about the performance improvement, I wasn't able to get the same numbers locally because there was too much variance. @JureCindro was this done on a per spec basis or for the whole suite? I'd love a 10% reduction of test execution time for nearly free 🙈

d4be4st:
one question remains:
bin/build and bin/deploy
vs
bin/build.sh and bin/deploy.sh

rest of the bin folder does not have an extension (the rails created ones) so i would lean towards the first option (without the extension). It looks prettier :D

uncoverd (approved):

ohaleck:
> one question remains:
> bin/build and bin/deploy
> vs
> bin/build.sh and bin/deploy.sh

According to the *nix contract, `bin` is expected to contain executables, with the `x` permission set, allowing them to be run without any auxiliary runtimes, interpreters etc. (like: `bash bin/build`). Whether they are actual binaries or scripts, is a secondary concern.
The function of file extension is the opposite.

Thus the extension should go.

On the other hand, if the environment is supposed to be strictly controlled (which is the usual thing in CI) and all commands were to be run in the form of `<interpreter> <script>` we could leave the scripts with extensions and provide extensionless symlinks to them.

template.rb
d4be4st: We can always do `puts config.seed` inside `rails_helper.rb` :)

But i will leave this for another time

PR #78, opened 2021-09-16
Overhaul of arch docs

d4be4st:
Task: no task

#### Aim
Simplify arch docs

#### Solution

Use links to terraform and devops wiki to simplify documentation


d4be4st:
Question that remains is: do we need a separate list of env variables?

Devops said it helped them on java projects, but i do not really know if that is really the case....

nikajukic (approved):
As for the question about env variables, I realized when I was doing it for loop that I basically c/p everything from secrets.yml. I think this list would become outdated soon so I would remove it. I don't see a benefit.

The devops said they're not going to use our tech docs either way, so :D

d4be4st:
Removed the env varibles file, renamed files :)

docs/architecture.md
```diff
@@ -0,0 +1,47 @@
...
+<!-- DEVELOPER -->
+<!-- if exists
+## Diagram
+
+![diagram](https://lucid.app/publicSegments/view/e1a4ca97-cf28-4b3b-8283-6e76a27f0158/image.png)
+-->
+
+## Aws Account
+
+* ACCOUNT-NAME (ACCOUNT-ID) <!-- infinum-dev (7021-9251-8610) --> <!-- DEVOPS -->
```
nikajukic, line 17: I just realized that this could be different per environment.

Can you please just change it so you can add multiple envs?

d4be4st: I though about it, but 99% of the time its a single AWS account

how about this solution? :)

nikajukic: Okay, then we're good to go! :)

PR #79, opened 2021-09-17
Add common env variables

d4be4st:
Task: [#353](https://app.productive.io/1-infinum/tasks/1669922)

#### Aim

To standardize our environment variable names a bit.

#### Solution

Add common environment variables

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -333,13 +333,28 @@ end
 SECRETS_YML_FILE = <<-HEREDOC.strip_heredoc
   default: &default
     secret_key_base: <%= Figaro.env.secret_key_base! %>
+
     database_name: <%= Figaro.env.database_name! %>
     database_username: <%= Figaro.env.database_username! %>
     database_password: <%= Figaro.env.database_password! %>
     database_host: <%= Figaro.env.database_host! %>
     database_port: <%= Figaro.env.database_port! %>
+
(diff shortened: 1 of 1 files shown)


d4be4st:
These are the env variables that appear most common in newer applications. Do you think i missed some?

template.rb
```diff
@@ -333,13 +333,28 @@
...
+    aws_s3_bucket: <%= Figaro.env.aws_s3_bucket! %>
+
+    mailgun_api_key: <%= Figaro.env.mailgun_api_key! %>
+    mailgun_domain: <%= Figaro.env.mailgun_domain! %>
+    mailgun_host: <%= Figaro.env.mailgun_host! %>
+
+    redis_url: <%= Figaro.env.redis_url! %>
+
+    # sidekiq_dashboard_username: <%= Figaro.env.sidekiq_dashboard_username! %>
+    # sidekiq_dashboard_password: <%= Figaro.env.sidekiq_dashboard_password! %>
```
nikajukic, line 357: Why is this commented out?

d4be4st: its 50/50 we are going to need this
If we have an admin dashboard we can use admin credentials to access the sidekiq webui
if there is no admin dashboard then we need the dedicated dashboard username/password

nikajukic (approved):

PR #85, opened 2021-11-11
Add slim-lint to template

lovro-bikic:
Task: [#402](https://app.productive.io/1-infinum/tasks/2039188)

#### Aim

Lint Slim files on `git push` to enforce a consistent style.

#### Solution

Add [slim-lint](https://github.com/sds/slim-lint) gem. Add a lint config file with an overriden rule.

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -301,6 +301,7 @@ append_to_file 'Gemfile' do
       gem 'brakeman', require: false
       gem 'bundler-audit', require: false
       gem 'rubocop-infinum', require: false
+      gem 'slim_lint', require: false
     end
   HEREDOC
 end
@@ -404,6 +405,9 @@ create_file 'config/application.yml', FIGARO_FILE
 # Rubocop
 get("#{BASE_URL}/.rubocop.yml", '.rubocop.yml')
 
+# Slim lint
+get("#{BASE_URL}/.slim-lint.yml", '.slim-lint.yml')
+
 # Mina
 get("#{BASE_URL}/mina_deploy.rb", 'config/deploy.rb')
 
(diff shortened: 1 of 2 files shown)


d4be4st (commented):
Should we add this to bin/build as well or should this be the next step once we are all ok with this?

template.rb
```diff
@@ -301,6 +301,7 @@
       gem 'brakeman', require: false
       gem 'bundler-audit', require: false
       gem 'rubocop-infinum', require: false
+      gem 'slim_lint', require: false
```
lovro-bikic, line 304: > Should we add this to bin/build as well or should this be the next step once we are all ok with this?
I thought of this at first, but I ran into a performance issue.

The project which I tested slim-lint on has 308 slim files. When I run
```
time bundle exec slim-lint app/views
```
this is the average result I get:
```
12.28s user 0.40s system 96% cpu 13.169 total
```
It takes a lot of time. (Note: the command isn't slow during `git push` because overcommit checks only changed files, and you typically have only a couple of those when you commit, so there's less to lint).

I tried to pinpoint what takes so much time to execute the command by excluding specific linters, and it turns out it's RuboCop. When I disable the RuboCop linter, like this:
```
time bundle exec slim-lint -x RuboCop app/views
```
this is the new average result:
```
1.45s user 0.27s system 73% cpu 2.333 total
```

So, I see three possible solutions:
1. Add slim-lint to bin/build with all linters enabled (the slowest but the most precise solution)
2. Add slim-lint to bin/build, but with RuboCop disabled (fast, but it comes with the cost of skipping RuboCop checks)
3. Don't add slim-lint to bin/build
What do you think? Personally, I would go with 1.

Take into consideration that specific lint times of course depend on the size of the project. For new projects the check won't be slow because there won't be that many Slim files, but as the codebase grows the check becomes slower.

(@d4be4st , I copy-pasted your question here so we can comment in a thread.)

d4be4st: And since we are moving to view components there slim files should be fewer and fewer :D

nikajukic (approved):

PR #89, opened 2021-12-20
Remove spring, pretty up the template

d4be4st:
Task: [#496](https://app.productive.io/1-infinum/tasks/2150453)

Remove spring from template
Pretty up the template

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -263,16 +263,17 @@ HEREDOC
 create_file 'config/initializers/bugsnag.rb', BUGSNAG_CONFIG
 
 # Remove gems we don't use.
-%w(coffee-rails jbuilder tzinfo-data byebug).each do |unwanted_gem|
-  gsub_file('Gemfile', /gem '#{unwanted_gem}'.*\n/, '')
+%w(jbuilder tzinfo-data byebug web-console).each do |unwanted_gem|
+  gsub_file('Gemfile', /gem "#{unwanted_gem}".*\n/, '')
 end
 
 # Remove comments from the Gemfile
 gsub_file('Gemfile', /^\s*#+.*\n/, '')
 
 # Add gems
-append_to_file 'Gemfile', after: /gem 'rails'.*\n/ do
+append_to_file 'Gemfile', after: /gem "rails".*\n/ do
   <<-HEREDOC.strip_heredoc
+
     gem 'bugsnag'
     gem 'figaro'
     gem 'pry-byebug'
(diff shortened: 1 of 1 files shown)


cilim (approved):

nikajukic (approved):

uncoverd (changes requested):
Did we do any benchmarking on the removal of spring? I know that on M1s and NLBL for example the difference isn't that big, but I remember it being substantial on other projects (maybe Loop) and on my older machine 🤔 

I wouldn't just remove spring based on DHH's will, he's got the latest iMac Pro and a large chunk of the backend team is still on intel based macs. I propose we do a real life test of console boot times and rspec runs on a few representative projects before we go forward with this and consider it the standard

cilim:
> Did we do any benchmarking on the removal of spring? I know that on M1s and NLBL for example the difference isn't that big, but I remember it being substantial on other projects (maybe Loop) and on my older machine 🤔
> 
> I wouldn't just remove spring based on DHH's will, he's got the latest iMac Pro and a large chunk of the backend team is still on intel based macs. I propose we do a real life test of console boot times and rspec runs on a few representative projects before we go forward with this and consider it the standard

Just bear in mind that a lot of people don't use spring, even if it's listed in the Gemfile and bundled on their machines 😄 
And even if they have slower machines than the M1

d4be4st:
> Did we do any benchmarking on the removal of spring? I know that on M1s and NLBL for example the difference isn't that big, but I remember it being substantial on other projects (maybe Loop) and on my older machine 🤔

No, but its kinda broken for `rails` commands anyway.
so `rake db:migrate` uses spring
but `rails db:migrate` does not

https://github.com/rails/spring#rails-console-rails-generate-rails-runner

anyway, if the project team wants spring (like if at one point in time rspec starting become just to long) they can add spring at a later point in the project.

d4be4st:
> Just bear in mind that a lot of people don't use spring, even if it's listed in the Gemfile and bundled on their machines 😄
And even if they have slower machines than the M1

By a lot people you mean you, right @cilim ? :D

cilim:
> > Just bear in mind that a lot of people don't use spring, even if it's listed in the Gemfile and bundled on their machines 😄
> > And even if they have slower machines than the M1
> 
> By a lot people you mean you, right @cilim ? :D

I'm the only one that **admits** I'm not using it, but there are more of us 😁

PR #93, opened 2022-01-03
Task #419 - GitHub Actions workflows

lovro-bikic:
Task: [#419](https://app.productive.io/1-infinum/tasks/2044901)

#### Aim

Replace Semaphore with GitHub Actions. 

#### Solution

Add reusable workflows for building and deploying. Add workflows which will be copied to the repo and which use the reusable workflows. Separate `bin/build` into `bin/lint` and `bin/test`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,107 @@
+name: Build
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  workflow_call:
+    inputs:
+      # Selects the version of Postgres for running tests
+      # See: https://github.com/docker-library/docs/blob/master/postgres/README.md#supported-tags-and-respective-dockerfile-links
+      postgres_image:
+        required: true
+        type: string
+
+      # Determines whether to install Node and run `yarn install`
+      use_node:
+        required: false
+        type: boolean
+        default: true
+
+      # Sets BUNDLE_APP_CONFIG environment variable
+      # See: https://bundler.io/man/bundle-config.1.html
+      bundle_app_config:
+        required: false
+        type: string
+        default: .bundle/ci-build
+
+      # Selects the runner on which the workflow will run
+      # See: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
+      runner:
+        required: false
+        type: string
+        default: ubuntu-20.04
+
+      # Defines which scripts will run on CI
+      # Format: space-delimited paths to scripts
+      # Example: 'bin/audit bin/lint bin/test'
+      ci_steps:
+        required: true
+        type: string
+    secrets:
+      VAULT_ADDR:
+        required: true
+      VAULT_AUTH_METHOD:
+        required: true
+      VAULT_AUTH_USER_ID:
+        required: true
+      VAULT_AUTH_APP_ID:
+        required: true
+
+jobs:
+  build:
+    name: 'Build'
+    runs-on: ${{ inputs.runner }}
+    env:
+      BUNDLE_APP_CONFIG: ${{ inputs.bundle_app_config }}
+      RUBOCOP_CACHE_ROOT: .rubocop-cache
+    services:
+      postgres:
+        image: postgres:${{ inputs.postgres_image }}
+        env:
+          POSTGRES_HOST_AUTH_METHOD: trust
+        ports:
+          - 5432:5432
+        options: --name=postgres
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+      - name: Prepare RuboCop cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ env.RUBOCOP_CACHE_ROOT }}
+          key: ${{ runner.os }}-rubocop-cache-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-rubocop-cache-
+      - name: Set up Node
+        uses: actions/setup-node@v2
+        if: ${{ inputs.use_node }}
+        with:
+          node-version-file: '.node-version'
+      - name: Prepare node_modules cache
+        uses: actions/cache@v2
+        if: ${{ inputs.use_node }}
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-modules-
+      - name: Install JS packages
+        if: ${{ inputs.use_node }}
+        run: yarn install --frozen-lockfile
+      - name: Prepare CI
+        run: bin/prepare_ci
+        env:
+          VAULT_ADDR: ${{ secrets.VAULT_ADDR }}
+          VAULT_AUTH_METHOD: ${{ secrets.VAULT_AUTH_METHOD }}
+          VAULT_AUTH_USER_ID: ${{ secrets.VAULT_AUTH_USER_ID }}
+          VAULT_AUTH_APP_ID: ${{ secrets.VAULT_AUTH_APP_ID }}
+      - name: Wait for Postgres to be ready
+        run: until docker exec postgres pg_isready; do sleep 1; done
+      - name: CI steps
+        run: 'parallel --lb -k -j0 ::: ${{ inputs.ci_steps }}'
--- a/README.md
+++ b/README.md
@@ -23,6 +23,22 @@ then run if needed:
   rbenv global #{latest_ruby}
 ```
 
+### GitHub Actions
+
+This template uses GitHub Actions for CI/CD. In order for workflows to work properly some [secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) have to be set up.
+
+For build workflow to work, the following secrets must exist (usually set up by DevOps):
+- `VAULT_ADDR`
+- `VAULT_AUTH_METHOD`
+- `VAULT_AUTH_USER_ID`
+- `VAULT_AUTH_APP_ID`
+
+For deploy workflows, you need to generate private/public SSH key pairs for each environment. Public key should be added to the server to which you're deploying. Private key should be added as a secret to GitHub and named `SSH_PRIVATE_KEY_#{ENVIRONMENT}`, where `ENVIRONMENT` is replaced with an appropriate environment name (`STAGING`, `PRODUCTION`, etc.).
+
+### Frontend
+
+If your application will have a frontend (the template will ask you that), you must have Node installed on your machine. The template creates a `.node-version` file with the Node version set to the version you're currently running (check by executing `node -v`). Therefore, ensure that you have the latest [Active LTS](https://nodejs.org/en/about/releases/) version of Node running on your machine before using the template.
+
 ## Usage
 
 ```shell
--- a/build.yml
+++ b/build.yml
@@ -1,0 +1,17 @@
+name: Build
+
+on: [push]
+
+jobs:
+  build:
+    name: Build
+    uses: infinum/default_rails_template/.github/workflows/build.yml@v1
+    with:
+      postgres_image: '13.2'
+      use_node: false
+      ci_steps: 'bin/audit bin/lint bin/test'
+    secrets:
+      VAULT_ADDR: ${{ secrets.VAULT_ADDR }}
+      VAULT_AUTH_METHOD: ${{ secrets.VAULT_AUTH_METHOD }}
+      VAULT_AUTH_USER_ID: ${{ secrets.VAULT_AUTH_USER_ID }}
+      VAULT_AUTH_APP_ID: ${{ secrets.VAULT_AUTH_APP_ID }}
--- a/deploy-production.yml
+++ b/deploy-production.yml
@@ -1,0 +1,19 @@
+name: Deploy production
+
+on:
+  workflow_dispatch:
+  # workflow_run: # UNCOMMENT THIS IF YOU WANT AUTOMATIC PRODUCTION DEPLOYS
+  #   workflows: [Build]
+  #   branches: [master]
+  #   types: [completed]
+
+jobs:
+  deploy:
+    name: Deploy
+    uses: infinum/default_rails_template/.github/workflows/deploy.yml@v1
+    with:
+      environment: production
+      branch: master
+      deployers: 'DEPLOY USERS GO HERE' # Example: '@github_username1 @github_username2'
+    secrets:
+      SSH_PRIVATE_KEY: ${{ secrets.SSH_PRIVATE_KEY_PRODUCTION }}
--- a/deploy-staging.yml
+++ b/deploy-staging.yml
@@ -1,0 +1,19 @@
+name: Deploy staging
+
+on:
+  workflow_dispatch:
+  workflow_run:
+    workflows: [Build]
+    branches: [staging]
+    types: [completed]
+
+jobs:
+  deploy:
+    name: Deploy
+    uses: infinum/default_rails_template/.github/workflows/deploy.yml@v1
+    with:
+      environment: staging
+      branch: staging
+      deployers: 'DEPLOY USERS GO HERE' # Example: '@github_username1 @github_username2'
+    secrets:
+      SSH_PRIVATE_KEY: ${{ secrets.SSH_PRIVATE_KEY_STAGING }}
--- a/template.rb
+++ b/template.rb
@@ -135,55 +135,67 @@ BIN_UPDATE = <<-HEREDOC.strip_heredoc
 HEREDOC
 create_file 'bin/update', BIN_UPDATE, force: true
 
-BIN_BUILD = <<~HEREDOC.strip_heredoc
+BIN_PREPARE_CI = <<~HEREDOC.strip_heredoc
   #!/usr/bin/env bash
 
   set -o errexit
   set -o pipefail
   set -o nounset
 
-  echo "=========== setting env variables ==========="
-  export RAILS_ENV='test'
-  export BUNDLE_APP_CONFIG='.bundle/ci-build'
+  echo "=========== pull secrets ==========="
+  bundle exec secrets pull -e development -y
+HEREDOC
+create_file 'bin/prepare_ci', BIN_PREPARE_CI, force: true
+chmod 'bin/prepare_ci', 0755, verbose: false
 
-  echo "=========== install bundler ==========="
-  bundler_version=`tail -n 1 Gemfile.lock | xargs`
-  time gem install bundler:$bundler_version --no-document
+BIN_AUDIT = <<~HEREDOC.strip_heredoc
+  #!/usr/bin/env bash
 
-  echo "=========== bundle install ==========="
-  time bundle install
-
-  echo "=========== secrets pull ============="
-  time bundle exec secrets pull -e development -y
-
-  echo "=========== rails db:test:prepare ==========="
-  time bundle exec rails db:test:prepare
+  set -o errexit
+  set -o pipefail
+  set -o nounset
 
   echo "=========== bundle audit ==========="
   time bundle exec bundle-audit update --quiet
   time bundle exec bundle-audit check
 
-  #############################################
-  # Uncomment this if you need yarn libraries #
-  # for running your tests                    #
-  #############################################
-  # echo "=========== yarn install ==========="
-  # time yarn install
+  echo "=========== brakeman ==========="
+  time bundle exec brakeman -q --color
+HEREDOC
+create_file 'bin/audit', BIN_AUDIT, force: true
+chmod 'bin/audit', 0755, verbose: false
+
+BIN_LINT = <<~HEREDOC.strip_heredoc
+  #!/usr/bin/env bash
+
+  set -o errexit
+  set -o pipefail
+  set -o nounset
 
   echo "=========== zeitwerk check ==========="
   time bundle exec rails zeitwerk:check
 
-  echo "=========== brakeman ==========="
-  time bundle exec brakeman -q
-
   echo "=========== rubocop  ==========="
-  time bundle exec rubocop --format simple
+  time bundle exec rubocop --format simple --format github --color --parallel
+HEREDOC
+create_file 'bin/lint', BIN_LINT, force: true
+chmod 'bin/lint', 0755, verbose: false
+
+BIN_TEST = <<~HEREDOC.strip_heredoc
+  #!/usr/bin/env bash
+
+  set -o errexit
+  set -o pipefail
+  set -o nounset
+
+  echo "=========== rails db:test:prepare ==========="
+  time RAILS_ENV=test bundle exec rails db:test:prepare
 
   echo "=========== rspec ==========="
-  time bundle exec rspec
+  time bundle exec rspec --force-color
 HEREDOC
-create_file 'bin/build', BIN_BUILD, force: true
-chmod 'bin/build', 0755, verbose: false
+create_file 'bin/test', BIN_TEST, force: true
+chmod 'bin/test', 0755, verbose: false
 
 BIN_DEPLOY = <<~HEREDOC.strip_heredoc
   #!/usr/bin/env bash
@@ -194,18 +206,11 @@ BIN_DEPLOY = <<~HEREDOC.strip_heredoc
 
   echo "=========== setting env variables ==========="
   environment=$1
-  export RAILS_ENV='test'
-  export BUNDLE_APP_CONFIG='.bundle/ci-deploy'
-
-  echo "=========== install bundler ==========="
-  bundler_version=`tail -n 1 Gemfile.lock | xargs`
-  time gem install bundler:$bundler_version --no-document
-
-  echo "=========== bundle install ==========="
-  time bundle install
 
   echo "=========== mina deploy =============="
-  time bundle exec mina $environment ssh_keyscan_domain setup deploy
+  time bundle exec mina $environment ssh_keyscan_domain
+  time bundle exec mina $environment setup
+  time bundle exec mina $environment deploy
 
   #############################################
   # Uncomment this if you need to publish dox #
@@ -227,6 +232,10 @@ BIN_DEPLOY = <<~HEREDOC.strip_heredoc
 HEREDOC
 create_file 'bin/deploy', BIN_DEPLOY, force: true
 chmod 'bin/deploy', 0755, verbose: false
+
+# get("#{BASE_URL}/build.yml", '.github/workflows/build.yml')
+# get("#{BASE_URL}/deploy-staging.yml", '.github/workflows/deploy-staging.yml')
+# get("#{BASE_URL}/deploy-production.yml", '.github/workflows/deploy-production.yml')
 
 # bundler config
 BUNDLER_CI_BUILD_CONFIG = <<~HEREDOC.strip_heredoc
@@ -578,6 +587,10 @@ if yes?('Will this application have a frontend? [No]', :green)
 
   get("#{BASE_URL}/.slim-lint.yml", '.slim-lint.yml')
 
+  node_version = `node -v`.chomp.sub('v', '')
+
+  create_file '.node-version', node_version
+
   PACKAGE_JSON_FILE = <<~HEREDOC
     {
       "name": "#{app_name}",
@@ -592,6 +605,9 @@ if yes?('Will this application have a frontend? [No]', :green)
       },
       "eslintConfig": {
         "extends": "@infinumrails/eslint-config-js"
+      },
+      "engines": {
+        "node": "#{node_version}"
       }
     }
   HEREDOC
@@ -634,7 +650,7 @@ if yes?('Will this application have a frontend? [No]', :green)
 
   create_file '.stylelintignore', STYLELINTIGNORE_FILE
 
-  append_to_file 'bin/build', after: "time bundle exec rubocop --format simple\n" do
+  append_to_file 'bin/lint' do
     <<~HEREDOC
 
       echo "=========== slim lint ==========="
@@ -649,6 +665,8 @@ if yes?('Will this application have a frontend? [No]', :green)
   end
 
   run 'yarn add --dev @infinumrails/eslint-config-js @infinumrails/stylelint-config-scss eslint postcss stylelint'
+
+  gsub_file('.github/workflows/build.yml', /^.*use_node: false.*\n/, '')
 end
 
 ## Ask about default PR reviewers
@@ -658,6 +676,13 @@ append_to_file '.github/CODEOWNERS' do
   * #{default_reviewers}
   HEREDOC
 end
+
+## Users allowed to manually trigger deploys
+staging_deployers = ask('Who can manually trigger a deploy to staging? (Example: @username1 @username2)', :green)
+gsub_file('.github/workflows/deploy-staging.yml', 'DEPLOY USERS GO HERE', staging_deployers)
+
+production_deployers = ask('Who can manually trigger a deploy to production? (Example: @username1 @username2)', :green)
+gsub_file('.github/workflows/deploy-production.yml', 'DEPLOY USERS GO HERE', production_deployers)
 
 # add annotate task file and ignore its rubocop violations
 rails_command 'generate annotate:install'
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,62 @@
+name: Deploy
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+
+on:
+  workflow_call:
+    inputs:
+      # Sets the Mina environment (e.g. staging, production)
+      # A task by the same name must exist in config/deploy.rb
+      environment:
+        required: true
+        type: string
+
+      # Sets the Git branch which will be checked out
+      branch:
+        required: true
+        type: string
+
+      # Determines who can manually trigger the workflow
+      # Example: "@github_username1 @github_username2"
+      # See: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
+      deployers:
+        required: false
+        type: string
+        default: ''
+
+      # Sets BUNDLE_APP_CONFIG environment variable
+      # See: https://bundler.io/man/bundle-config.1.html
+      bundle_app_config:
+        required: false
+        type: string
+        default: .bundle/ci-deploy
+
+      # Selects the runner on which the workflow will run
+      # See: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
+      runner:
+        required: false
+        type: string
+        default: ubuntu-20.04
+    secrets:
+      SSH_PRIVATE_KEY:
+        required: true
+
+jobs:
+  deploy:
+    name: Deploy
+    runs-on: ${{ inputs.runner }}
+    env:
+      BUNDLE_APP_CONFIG: ${{ inputs.bundle_app_config }}
+    if: ${{ github.event_name == 'workflow_dispatch' && contains(inputs.deployers, format('@{0}', github.actor)) || github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ inputs.branch }}
+      - uses: ruby/setup-ruby@v1
+        with:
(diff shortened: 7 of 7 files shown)


workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+      BUNDLE_APP_CONFIG: ${{ inputs.bundle_app_config }}
+    services:
+      postgres:
+        image: ${{ inputs.postgres_image }}
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+        ports:
+          - 5432:5432
+        options: --name=postgres
```
lovro-bikic, line 45: This is a Postgres service which will be used to run tests. There is nothing that special here (I pretty much followed best practices for setting this up), except a _tiny_ difference, which will also explain why I split `bin/build` into `bin/lint` and `bin/test`.

[Typically](https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers#running-jobs-in-containers), Postgres is set up with the following options:
```yml
# Set health checks to wait until postgres has started
options: >-
  --health-cmd pg_isready
  --health-interval 10s
  --health-timeout 5s
  --health-retries 5
``` 

When you start a Postgres docker image, Postgres service usually isn't available immediately. Therefore, if one would run `docker run postgres` and immediately after that run `bundle exec rspec`, it would most probably fail because even though the container is running, the Postgres service wouldn't be ready at the time of executing tests.

This problem is usually solved by supplying a health check command which is run every couple of seconds until Postgres becomes ready (the above snippet). The process looks approximately like this:
- Start the container
- Run the `pg_isready` command, the container replies that Postgres isn't ready
- Sleep X seconds (The GA workflow is essentially blocked for X seconds)
- Run the `pg_isready` command, the container replies that Postgres isn't ready
- Sleep Y seconds
- Run the `pg_isready` command, the container replies that Postgres isn't ready
- Sleep Z seconds
- Run the `pg_isready` command, the container is ready!
- Run tests and everything else

The problem with this approach is having to wait all that time because `sleep` is invoked and nothing happens in the meantime (You can see an example of a workflow run with this approach [here](https://github.com/infinum/rails-infinum-revisor/runs/4622326769?check_suite_focus=true), the step you're looking for is `Initialize containers`). It takes 26-27 seconds on average to get Postgres up and running with this approach.

However, I opted for a different approach which saves us some time (remember that GA workflow runs are billed per minute, so time optimizations can be cost-effective in the long run).

This is my approach. We actually don't need Postgres until the point in the workflow in which tests are run, and this happens at the very end. In between initializing the Postgres service and actually connecting to Postgres, a lot happens (Ruby is set up, gems are installed, Node is set up, JS packages are installed, secrets are pulled, linters are run).

So, instead of waiting on Postgres to be ready, I only start the container and immediately start executing everything else. The Postgres service is still being prepared in the background while everything else is happening (those two things happen in parallel).

When everything else that doesn't require Postgres executes, and when we reach the point of needing Postgres, I then run this command: `until docker exec postgres pg_isready; do sleep 1; done`. This will run the same health check command `pg_isready` as in the first approach, and will also sleep if the service isn't ready yet. However, because of all the Ruby and Node setup, secrets pulling and linters running took some time to execute, by the time they're all done, Postgres will already be ready. In practice, `sleep 1` will never even execute, and tests will automatically follow after this command.

An example workflow run with this approach is [here](https://github.com/infinum/rails-infinum-revisor/runs/4688854144?check_suite_focus=true). This way, Postgres start up on average takes less than 10 seconds.

The latter approach is ~16 seconds faster than the original approach. I hope this will save us some billable time.

However, the faster approach isn't possible if we don't split `bin/build` into two files:
- part that doesn't depend on Postgres -> `bin/lint`
- part that depends on Postgres -> `bin/test`

The `bin/lint` part of the workflow also gives us some "free" time to wait for Postgres to be ready. I think it also conceptually makes sense to have linting and testing as separate scripts.

workflows/deploy.yml
```diff
@@ -0,0 +1,45 @@
+name: Deploy
```
lovro-bikic, line 1: This is a reusable workflow, same as the build one above.

workflows/deploy.yml
```diff
@@ -0,0 +1,45 @@
...
+        required: true
+        type: string
+      deployers:
+        required: false
+        type: string
+        default: ''
+      bundle_app_config:
+        required: false
+        type: string
+        default: .bundle/ci-deploy
```
lovro-bikic, line 23: These parameters must be provided by the workflow which references this reusable workflow.

`environment` is the same environment used for Mina (e.g. `production`).

`branch` is the git branch which will be checked out.

`deployers` is a list of people (their GH usernames) who are allowed to trigger a deploy via the UI. It can look like this:
```yml
deployers: '@lovro-bikic @melcha'
```
Anyone not in this list won't be allowed to deploy manually from UI. Note: everyone with write access to the repo can run the workflow from the UI, however the workflow will be stopped if they're not in the list.

`bundle_app_config` was explained above.

workflows/deploy.yml
```diff
@@ -0,0 +1,45 @@
...
+    secrets:
+      SSH_PRIVATE_KEY:
+        required: true
```
lovro-bikic, line 26: This is a private SSH key which must be provided in order for SSH to work with the server (the server will have the public key).

workflows/deploy.yml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-20.04
+    env:
+      BUNDLE_APP_CONFIG: ${{ inputs.bundle_app_config }}
+    if: ${{ github.event_name == 'workflow_dispatch' && contains(inputs.deployers, format('@{0}', github.actor)) || github.event.workflow_run.conclusion == 'success' }}
```
lovro-bikic, line 34: This is the condition under which the workflow is run. There are two cases: `workflow_dispatch` and `workflow_run`.

`workflow_dispatch` means manual trigger. In this case we check who the allowed deployers are (`contains(inputs.deployers, format('@{0}', github.actor))`).

`workflow_run` happens when a Build workflow run has completed. In this case we check if that run was successful (`github.event.workflow_run.conclusion == 'success'`).

deploy-production.yml
```diff
@@ -0,0 +1,15 @@
+name: Deploy production
+
+on:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
d4be4st: I do not think that is what you wrote.
The branch selection here is where do you want the workflow to be used from, not which branch is going to be "deployed".

At least that is how I understood it.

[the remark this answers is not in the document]
d4be4st: arent you missing the `workflow_run` here as you do in staging?

template.rb
```diff
@@ -680,6 +666,13 @@ def run
   HEREDOC
 end
 
+## Users allowed to manually trigger deploys
+staging_deployers = ask('Who can manually trigger a deploy to staging? (Example: @username1 @username2)', :green)
+gsub_file('.github/workflows/deploy-staging.yml', 'DEPLOY USERS GO HERE', staging_deployers)
```
d4be4st, line 671: do we want this?

I see the appeal, but it might just be a hassle in the end...

How many people are distraught because they can rebuild/redeploy things in semaphore?


Also its easy to change the .yml file for a user and just push, no? :)

workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+        default: .bundle/ci-build
+    secrets:
+      VAULT_ADDR:
+        required: true
+      VAULT_AUTH_METHOD:
+        required: true
+      VAULT_AUTH_USER_ID:
+        required: true
+      VAULT_AUTH_APP_ID:
+        required: true
```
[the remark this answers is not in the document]
d4be4st: here is a question to think about:

Do we maybe want to change this us? Istead of pulling from vault just add a signle secret as the contents of the whole application.yml file?

and have a `run: echo $SECRETS > config/application.yml` as a job?


Pros: 
We reduce the need for GHA to have access to our vault
One less user/app creds on vault for devops to create

Cons:
you need an admin to update the secrets

workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+          POSTGRES_PASSWORD: postgres
+        ports:
+          - 5432:5432
+        options: --name=postgres
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+      - uses: actions/setup-node@v2
```
d4be4st, line 51: so we need 3 steps for node while ruby has that all in one step? 

🤦

workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: bundle exec secrets pull -e development -y
+        env:
+          VAULT_ADDR: ${{ secrets.VAULT_ADDR }}
+          VAULT_AUTH_METHOD: ${{ secrets.VAULT_AUTH_METHOD }}
+          VAULT_AUTH_USER_ID: ${{ secrets.VAULT_AUTH_USER_ID }}
+          VAULT_AUTH_APP_ID: ${{ secrets.VAULT_AUTH_APP_ID }}
+      - name: Install JS packages
+        if: ${{ inputs.use_node }}
+        run: yarn install
+      - name: Run linters
```
d4be4st, line 72: in one of my previous ventures i setup the linters to run paralel to the specs
so each linter is run independently and once all have been run then the specs are run

it might save a couple of real world minutes because its parallel :)

d4be4st (commented):
Have you tested this out on any project?
I am wondering how the slacks /github integration has progressed.

Previously I had to use some slack/notifier actions to report to the slack which was not the best to be honest :D

template.rb
lovro-bikic: > Also its easy to change the .yml file for a user and just push, no?

Not if the master branch is protected (and we typically have that now), so even if someone wants to change the list they first have to get approval from someone with master access (e.g. lead dev).

But as far as I know, not everyone with write access is automatically allowed to deploy to production, that's why I added this.

deploy-production.yml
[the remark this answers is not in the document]
lovro-bikic: > arent you missing the workflow_run here as you do in staging?

Nope, because deploying to production should only be done manually (by default). If someone wants to automatically deploy to production on push to master, they can add the `workflow_run` event.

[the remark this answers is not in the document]
lovro-bikic: > The branch selection here is where do you want the workflow to be used from, not which branch is going to be "deployed".

From the [docs](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#workflow_dispatch):
image

So if you select `staging` branch for example and then use `actions/checkout`, it will check out `staging`.

workflows/build.yml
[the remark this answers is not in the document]
lovro-bikic: I'm not sure, but the immediate problem that I see is having to rely on two sources of truth for secrets. One source of truth is Vault, the other is GitHub. If you update one but not the other, you'll run into problems (and people are forgetful, I would most probably forget to update secrets on GH). I think it's better that we have a single source of truth.

workflows/build.yml
lovro-bikic: Yes, because `setup-node` actions doesn't run `yarn install` (`setup-ruby` on the other hand runs `bundle install`), and I need to cache node modules (`setup-node` can easily miss cache so I had to set it up manually (see [here](https://github.com/actions/setup-node/issues/328))).

lovro-bikic:
> Have you tested this out on any project?

@d4be4st it's currently live on [Revisor](https://github.com/infinum/rails-infinum-revisor/actions).

deploy-production.yml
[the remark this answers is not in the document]
d4be4st: > Nope, because deploying to production should only be done manually (by default). If someone wants to automatically deploy to production on push to master, they can add the workflow_run event.

Hm... this is something we all need to agree on as its a major shift from what used to be. (Or per project base)
I vote for an automatic deploy as most code ending in master comes from PR merges.

nikajukic:
@melcha @cilim @JureCindro @uncoverd @vr4b4c please take a look at this PR

deploy-staging.yml
```diff
@@ -0,0 +1,19 @@
...
+    types: [completed]
+
+jobs:
+  deploy:
+    name: Deploy
+    uses: infinum/default_rails_template/workflows/deploy.yml@v1
+    with:
+      environment: staging
+      branch: staging
+      deployers: 'DEPLOY USERS GO HERE'
```
nikajukic, line 17: Maybe write the example in the comment for what is expected here:
`'@github_username_1 @github_username_2'`

Same for deploy-production.yml

deploy-production.yml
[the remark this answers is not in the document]
nikajukic: I think it should be configurable per project. 

On Loop I don't have automatic deploys at all because the deploy takes around 15mins and I usually bundle a few PRs together and then deploy them. 

On most other projects I don't have automatic deploy to production.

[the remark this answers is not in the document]
nikajukic: @lovro-bikic can you add `workflow_run` event and comment it out with instructions on when/how to use it?

nikajukic (approved):

build.yml
```diff
@@ -0,0 +1,16 @@
+name: Build
+
+on: [push]
+
+jobs:
+  build:
+    name: Build
+    uses: infinum/default_rails_template/workflows/build.yml@v1
+    with:
+      postgres_image: '14.1'
```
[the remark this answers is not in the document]
cilim: Is the postgres port by default 5432:5432? In case it's not, maybe write it as:

```suggestion
      postgres:
        image: postgres:14.1
```

Also, don't we need to setup some environment variables for this? For example, the postgres password or database name?

deploy-production.yml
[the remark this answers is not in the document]
cilim: I was also under the impression that all of our projects have automatic production deploys. Guess we don't 🤷 😄 

I'm okay with changing this workflow for all, but it would be cool to make it optional, so project teams can set this up as they want.

workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+  lint_and_test:
+    name: 'Lint and test'
+    runs-on: ubuntu-20.04
+    env:
+      BUNDLE_APP_CONFIG: ${{ inputs.bundle_app_config }}
+    services:
+      postgres:
+        image: postgres:${{ inputs.postgres_image }}
+        env:
+          POSTGRES_USER: postgres
```
cilim, line 41: ah, this is setup here, got it! 👌

deploy-production.yml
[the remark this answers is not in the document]
lovro-bikic: > can you add workflow_run event and comment it out with instructions on when/how to use it?

Added.

README.md
```diff
@@ -23,6 +23,18 @@ then run if needed:
...
+
+This template uses GitHub Actions for CI/CD. In order for workflows to work properly some [secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) have to be set up.
+
+For build workflow to work, the following secrets must exist (usually set up by DevOps):
+- `VAULT_ADDR`
+- `VAULT_AUTH_METHOD`
+- `VAULT_AUTH_USER_ID`
+- `VAULT_AUTH_APP_ID`
+
+For deploy workflows, you need to generate private/public SSH key pairs for each environment. Public key should be added to the server to which you're deploying. Private key should be added as a secret to GitHub and named `ENVIRONMENT_KEY`, where `ENVIRONMENT` is replaced with an appropriate environment name (`STAGING`, `PRODUCTION`, etc.).
```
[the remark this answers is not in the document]
lovro-bikic: Changed.

build.yml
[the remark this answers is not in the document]
lovro-bikic: > We don't have postgres 14.1 yet I think. Please check with devops. I think the latest RDS version we have is 13.2

Yep, RDS still doesn't offer version 14. I changed it back to 13.2.

workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Pull secrets
+        run: bundle exec secrets pull -e development -y
+        env:
+          VAULT_ADDR: ${{ secrets.VAULT_ADDR }}
+          VAULT_AUTH_METHOD: ${{ secrets.VAULT_AUTH_METHOD }}
+          VAULT_AUTH_USER_ID: ${{ secrets.VAULT_AUTH_USER_ID }}
+          VAULT_AUTH_APP_ID: ${{ secrets.VAULT_AUTH_APP_ID }}
+      - name: Install JS packages
+        if: ${{ inputs.use_node }}
+        run: yarn install
```
uncoverd, line 71: I'm not sure exactly how it is with yarn, do we have to be careful here like we are with `npm install` vs `npm ci` ?
Ref: https://docs.npmjs.com/cli/v8/commands/npm-ci

workflows/build.yml
uncoverd: Screenshot 2022-01-10 at 11 03 48

Agreed, I think we care much more about the real world minutes then the billed minutes, our hour is not cheap 😄

workflows/deploy.yml
```diff
@@ -0,0 +1,45 @@
+name: Deploy
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
uncoverd, line 5: Should we really cancel in progress deploys here? That might get us into weird deployment states which could be tricky to recover from 🤔

workflows/deploy.yml
uncoverd: This comment sounds like great documentation, can we include it in the workflow for future reference🙏

README.md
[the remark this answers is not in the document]
lovro-bikic: @uncoverd This name was chosen in coordination with the Devops ([Slack ref](https://infinum.slack.com/archives/C0AME3E4Q/p1641294254126600)).

workflows/build.yml
lovro-bikic: Good call, this behavior can be achieved in yarn with the `--frozen-lockfile` flag which I've now added.

workflows/deploy.yml
lovro-bikic: True, I removed `cancel-in-progress`. I left the `group` part of the concurrency setup, so this will at least cancel pending jobs if there is more than one of them.

workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+        required: true
+      VAULT_AUTH_USER_ID:
+        required: true
+      VAULT_AUTH_APP_ID:
+        required: true
+
+jobs:
+  lint_and_test:
+    name: 'Lint and test'
+    runs-on: ubuntu-20.04
```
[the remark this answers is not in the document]
lovro-bikic: I can add a variable for this (with the default set to `ubuntu-20.04`), but what other runner would you put here? There's Windows, Mac, and an older version of Ubuntu.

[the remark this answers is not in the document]
uncoverd: 22.04 in April 🤞 

Screenshot 2022-01-10 at 12 12 33

README.md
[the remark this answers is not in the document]
ivantomica: When you phrase it like that, yes :-D Perhaps we need to change this on the WP side of things as well, to be on the same page.

As a favor towards DevOps team, perhaps you could say during setup process "we expect this to have ENVIRONMENT_SSH_KEY" or "ENVIRONMENT_DEPLOY_KEY" in environment variables on GHA"

workflows/build.yml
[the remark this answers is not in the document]
lovro-bikic: I added an optional input for a runner with the default set to `ubuntu-20.04`.

Just a note, the idea with reusable workflows is that at one point when a new runner becomes available (e.g. Ubuntu 22.04), we'll only have to update the runner in the reusable workflow, and all projects will be using Ubuntu 22 as well immediately (if it's not a breaking change, of course).

README.md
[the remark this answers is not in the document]
ivantomica: Checked with WP team, they're fine with the change. My hunch is telling me to go with ENVIRONMENT_DEPLOY_KEY. What do you think? Have preference one way or the other?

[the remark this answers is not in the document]
lovro-bikic: I'd go with `SSH_PRIVATE_KEY_#{ENVIRONMENT}`, it's the most accurate description of the secret.

workflows/build.yml
lovro-bikic: I can separate the workflow into separate jobs, but I have some concerns.

First, pricing. GHA charges each job separately. In the above example, `bundle-cache` is charged 1 billable minute, `specs` is charged 4 billable minutes, `lint` is charged 2 billable minutes, and `audit` is charged 1 billable minute. The total is 8 billable minutes.

Each of those jobs invokes the checkout action, and setup-ruby action. This typically takes 10 seconds (if we assume that no new gems are downloaded, i.e. everything is pulled from cache).

Compare this to a single job which does the checkout, Ruby setup, audit, specs, and linters. With the above numbers, it would approximately take:
10s (checkout, setup) + 2m 50s (specs - checkout, setup) + 1m 20s (lint - checkout, setup) + 10s (audit - checkout, setup) = 4m 30s

This is 5 billable minutes. The time difference is 3 billable minutes, and the cost difference is $0,024 (Linux costs $0,008 per billable minute). In itself, that's not costly. However, these jobs will be run hundreds of times on tens of projects. Suddenly, 3 billable minutes become 3000 billable minutes. And costs aggregate with time, so at the end of the year, the cost difference becomes even greater.

I generally don't have a problem with GHA costing more if that's appropriate for our use-case, but I just want everyone to be on the same page about the cost aspect.

My second concern is, why would we want to run specs if linting and/or audit fails? If any one of those jobs fails, the workflow run is considered failed.

vr4b4c (approved):

vr4b4c:
👏

workflows/build.yml
JureCindro: So we can compare the build times, but include the price of the developer in it. I think it is a safe assumption that during the build process, the developer is not doing anything other than waiting, on average. 
I know some (hyper-productive) people want to say that they reply on slack, reply on emails, etc during the build process. But often those tasks take more than the build time so then the merge is waiting. Maybe even someone else merges while the busy-body developer is replying to his mails and then it requires him to rebase the branch before merging again. I for one am so small minded that if I do anything else (productive at least) in between, I completely forget about the branch build.

JureCindro: So we can compare the build times, but include the price of the developer in it. I think it is a safe assumption that during the build process, the developer is not doing anything other than waiting, on average. 
I know some (hyper-productive) people want to say that they reply on slack, reply on emails, etc during the build process. But often those tasks take more than the build time so then the merge is waiting. Maybe even someone else merges while the busy-body developer is replying to his mails and then it requires him to rebase the branch before merging again. I for one am so small minded that if I do anything else (productive at least) in between, I completely forget about the branch build, and then merge half a day later when someone pings me regarding this.

So for sake of the exercise let's assume that the developer is waiting during build process that is required for the merge. I believe the billed hour is a bit more than 50 eur/h.
As you stated, the difference is 3 billable minutes. This saves $0,024 in container running time. The difference in developer billable minutes is thus `50 / 60 * 3` which is 2,5 €.
So in the example where we parallelise the build runs we save more than 100x the amount we pay more for the billable time of the runners.

For the second concern, I think it is beneficial to know that specs have completed successfully but only the audit failed. Same for any other combination. You'd also want to know if all of those fail, and not exit early if any of these 3 fail.

cilim (approved):
Awesome stuff! 👏

workflows/build.yml
d4be4st: To save time, can we introduce tags to run just specific jobs?
For example, rubocop linter failed because you had more than 120 line, and you want to fix it but you really just want to run the linter again

also [skip ci] tag would be handy to skip everything.

maybe something for the next iteration?

melcha: @lovro-bikic could checkout and ruby setup be extracted into a separate job, and then parallel lint, test and audit jobs would depend on the extracted setup job? 
Maybe we could save some minutes this way.
https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows#creating-dependent-jobs

template.rb
```diff
@@ -228,6 +210,10 @@
 create_file 'bin/deploy', BIN_DEPLOY, force: true
```
melcha, line 210: We might need to set `RAILS_ENV=test` for `db:test:prepare` explicitly.

Remember my PR from last week, I removed export RAILS_ENV=test from bin/deploy? I tested that on Cormeum, and the Semaphore deploy failed at this line because:

1. the default RAILS_ENV is development, if nothing is set
2. annotate gem is in development group in gemfile
3. we skip development group in `.bundle/ci-deploy/config`
4. lib/tasks/auto_annotate_models.rake checks if the rails env is development and then requires the gem and defines the task.

So `RAILS_ENV=test bundle exec rails db:test:prepare` works.

melcha (approved):
Excellent, @lovro-bikic !

template.rb
```diff
@@ -228,6 +210,10 @@
...
 chmod 'bin/deploy', 0755, verbose: false
 
+get("#{BASE_URL}/build.yml", '.github/workflows/build.yml')
```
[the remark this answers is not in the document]
lovro-bikic: This is at the top of the template:
```ruby
BASE_URL = 'https://raw.githubusercontent.com/infinum/default_rails_template/master'.freeze
```

template.rb
```diff
@@ -670,6 +656,8 @@ def run
   end
 
   run 'yarn add --dev @infinumrails/eslint-config-js @infinumrails/stylelint-config-scss eslint postcss stylelint'
+
+  gsub_file('.github/workflows/build.yml', /^.*use_node: false.*\n/, '')
```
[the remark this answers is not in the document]
lovro-bikic: So by default Node is turned off in the build workflow. This line is part of the conditional where the user is asked "Will this application have a frontend?", and if they respond yes, then Node installation is enabled by removing this line.

workflows/build.yml
lovro-bikic: > To save time, can we introduce tags to run just specific jobs?

Yes, this is possible. We can add this to the reusable workflow later on, and since it's backward-compatible, it can be added to v1.

build.yml
```diff
@@ -0,0 +1,16 @@
+name: Build
+
+on: [push]
+
+jobs:
+  build:
+    name: Build
+    uses: infinum/default_rails_template/workflows/build.yml@v1
```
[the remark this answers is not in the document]
lovro-bikic: @melcha can you set that up? I pushed the branch (it's currently on the same commit as master).

template.rb
lovro-bikic: I added it back.

build.yml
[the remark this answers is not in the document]
melcha: Done.

workflows/deploy.yml
lovro-bikic: Added in https://github.com/infinum/default_rails_template/pull/93/commits/71cc1741ea5f1407554c0f9599d06431fe902089.

JureCindro: Could we change the `deployers:` to accept an array which is common practice in yml or is there a special restriction in github actions that requires a string with space delimited username handles?

workflows/build.yml
cilim: To save time, maybe it would be good to run the lint and audit checks **only** when a branch is merged to the staging/master branch? 🤔

workflows/deploy.yml
lovro-bikic: Inputs support only booleans, numbers, and strings sadly ([docs](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onworkflow_callinputsinput_idtype)).

uncoverd (approved):
🙆‍♂️

workflows/build.yml
d4be4st: @cilim how are you going to "fix" the lint errors on master (protected branch) :D

melcha: In a separate PR? It shouldn't happen that often because one should have rubocop embedded into their text editor and see rubocop offenses when editing files, and run overcommit locally before each commit like we did until now. 
I honestly still think only zeitwerk out of these check must be a hard check on CI, rubocop and bundle audit should be ignorable.

ivantomica: Perhaps I'm going slightly of rails here, but my opinion was, and still is, we should not run Linting on each commit. It prolongs the build time with questionable usefulness and has the potential to cost us a lot.

What I would propose as an approach is:
- Linting - run locally using pre-commit/overcommit/whatever - also run on PR (I believe @cilim once mentioned that he first opens up PR and then starts developing in it - in my opinion - that's the wrong approach. If ABSOLUTELY necessary, alternative could be to explore possibility of ignoring Linting only on "Draft PR" feature Github has. And utilize that from the developer's end
- Security audit - should be part of the CI job on each PR to main branches - perhaps with option to merge regardless for the admins/TLs when situation requires it (:hankey: hits the fan and we need to fix something that's impairing production)
- Other checks - depends on what do they check :-) But we need to be sensible and ask ourselves "where do we get most value from this, while keeping cost in mind"

cilim: > I believe @cilim once mentioned that he first opens up PR and then starts developing in it 

fyi: no 😄 I mostly develop everything before even pushing to remote

uncoverd: I don't think we should be trading pennies for worse development outcomes. 

The only positive I see with reducing linting and bundle audit is that we save a few bucks. As @JureCindro said above, our development time is much more valuable than an external service's minute. In reality the test execution time takes the vast majority of the build time. The split we're seeing here is due to the team under @JureCindro watchful eye and their diligence to produce a well balanced test suite that has 100% coverage and runs extremely fast for the number of cases. Most other projects with at least decent coverage are much slower. In those cases linting and auditing becomes an even smaller portion of the bill.

Very few projects have been able to keep up with bundle audit and rubocop errors back in the day when these weren't mandatory on CI. I remember the old dance too well 

> "hey X, did you just push a change to staging that broke rubocop? God I hate you, I have to fix it in my staging merge now!
> "hey guys, did anyone notice that theres 5 bundle audit warnings?" "yeah I just ignored those, I needed to get 5 urgent fixes out"

@ivantomica I don't understand your argument for only running certain actions "on each PR", could you elaborate?
In general the CI pipeline runs when

- you push a branch for the first time (usually minutes before opening a PR)
- you push another commit to an existing branch (usually for PR fixes)
- you rebase the existing branch to the latest master (usually before merging)
- you merge a branch to master/uat/staging

I think we want the safety of the CI pipeline in all of these cases.
- When you first push a branch you want to see if its ready to be merged (auditing, linting, specs).
- Once you push a PR fix you again want to know if the change is safe (auditing, linting, specs).
- When you're just about ready to merge the feature branch to master you again want to be sure that the rebased version is still valid since its likely there were some code changes.
- Finally when you merge a feature branch into master you are constructing a new variation of the code that contains the original master + your feature branch changes and you again want to be sure the build passes as that code will get shipped.

I won't get too deep into the CI vs local debate again (we did this before, reference: https://github.com/infinum/rails-jsonapi-example-app/pull/42, https://github.com/infinum/default_rails_template/pull/63). All of these checks being done on each change automatically mean that the team fixes security issues earlier (since they break the build), stays in sync as far as code style is concerned and is confident that the code that is going out is always up to standards. This is in my opinion the cheapest win you can get, education and oversight is much more expensive.

@d4be4st I belive skip-ci already works out of the box, my gwip commits skip CI
Screenshot 2022-01-12 at 19 17 15

uncoverd: I forgot one more thing - @lovro-bikic mentioned today that rubocop is awfully slow on CI. Locally it takes ~7s while on GHA it takes around a minute usually. 

I was able to halve this using the --parallel flag and I'm betting theres more execution speed to be gained via some config options or profiling.
Screenshot 2022-01-12 at 20 14 19

.github/workflows/build.yml
```diff
@@ -0,0 +1,108 @@
...
+        required: true
+      VAULT_AUTH_METHOD:
+        required: true
+      VAULT_AUTH_USER_ID:
+        required: true
+      VAULT_AUTH_APP_ID:
+        required: true
+
+jobs:
+  build:
```
JureCindro, line 53: Should we add some default env variables here?

```yml
env:
  RAILS_ENV: test
  BUNDLE_WITHOUT: development
  ```

Comment (lovro-bikic) [reply .github/workflows/build.yml:53]: `BUNDLE_WITHOUT` is set through the `BUNDLE_APP_CONFIG` variable, which is defined per-project.

`RAILS_ENV` has been added where it's necessary (`bin/test`), but I don't see any harm in adding in globally.

**Comment by lovro-bikic** [comment]
There have been some changes since the last review, please take a look at the PR again.

Important changes:
- added `bin/audit` which has some of the steps from `bin/lint` (https://github.com/infinum/default_rails_template/pull/93/commits/ab144e3086fd0e97d1b53501a328c7ea7aa6ae26)
- added `bin/prepare_ci` instead of the hardcoded step for pulling secrets. I noticed that some projects have a more elaborate CI setup, so this gives greater flexibility to those projects (https://github.com/infinum/default_rails_template/pull/93/commits/5c2a87553b715ed11bfcb9d7cda2b97a9a7e6d73)
- steps were changed to run in parallel with the [parallel](https://www.gnu.org/software/parallel/) command. additionally, an input `ci_steps` was added so projects can define which files they want to run (by default, it's `bin/audit bin/lint bin/test`, but if a project has a different setup, they can override it).  (https://github.com/infinum/default_rails_template/pull/93/commits/5c2a87553b715ed11bfcb9d7cda2b97a9a7e6d73) (more info [here](https://infinum.slack.com/archives/C027RRR17/p1641996025020100))
- Added RuboCop cache. On one project, the cache reduced RuboCop execution time from 60s to 1s (https://github.com/infinum/default_rails_template/pull/93/commits/14869808ddb86b05ea5ab049a7ed9c3bfb7aa48a)

File: .github/workflows/build.yml [hunk 23]
```diff
@@ -0,0 +1,107 @@
...
+        required: false
+        type: string
+        default: ubuntu-20.04
+
+      # Defines which scripts will run on CI
+      # For the format, see the default value
+      ci_steps:
+        required: false
+        type: string
+        default: 'bin/audit bin/lint bin/test'
```
lovro-bikic, line 41: I wasn't sure whether it's better to set a default value here, or make the input required so that each project has to define in their own build workflow which steps to use. The second approach is more explicit, and developers can look into their project build.yml to figure out which scripts will run on CI. What do you think?

template.rb
```diff
@@ -135,55 +135,67 @@
...
 
   echo "=========== zeitwerk check ==========="
   time bundle exec rails zeitwerk:check
 
-  echo "=========== brakeman ==========="
-  time bundle exec brakeman -q
-
   echo "=========== rubocop  ==========="
-  time bundle exec rubocop --format simple
+  time bundle exec rubocop --format github --color --parallel
```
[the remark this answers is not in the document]
lovro-bikic: This is how it looks in the Files changed tab:
image

d4be4st (approved):

nikajukic (approved):

cilim (approved):

.github/workflows/build.yml
lovro-bikic: After a bit of thinking, I decided to make this input required and defined per-project. That way, developers can know how their CI will work just by looking at their `build.yml`. (https://github.com/infinum/default_rails_template/pull/93/commits/633753171a074db7d6c911d1417db44f857788d8)

JureCindro (approved):

uncoverd (approved):
🔥

workflows/build.yml
lovro-bikic: To summarize this whole thread: we've decided to run a single job, but run steps in parallel (using GNU parallel). This decision is a compromise: it's cheaper than the parallel jobs approach, but it still allows for auditing and linting on CI (for people who want that), without slowing down the single job.

Technically, this is achieved by splitting auditing, linting, and testing into three scripts, and calling them in parallel like this:
```
$ parallel --lb -k -j0 ::: bin/audit bin/lint bin/test
```
Flag explanations:
`--lb`: `line-buffer`, prints output line by line as it's produced (instead of waiting for the process to end before printing everything)
`-k`: `keep`, tells parallel not to mix output of different scripts (i.e. it will first output all lines from bin/audit, then all lines from bin/lint, and finally all lines from bin/test)
`-j0`: `jobs`, tells parallel to run as many jobs as possible in parallel. This will determined by the number of scripts parallel needs to run (by default it's 3, but some projects can use a different number of scripts)

PR #96, opened 2022-01-05
Remove bundler config folder from gitignore

melcha:
Task: n/a

Rails by default adds 
```
# Ignore bundler config.
/.bundle
```
to gitignore. Since we've started using this folder for bundle config instead of deprecated CLI arguments, we need to remove these lines from gitignore.

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -491,6 +491,9 @@ HEREDOC
 
 append_file '.gitignore', GITIGNORED_FILES
 
+# remove bundler ci config folders from gitignore
+append_to_file '.gitignore', "/*\n!/.bundle/ci-build\n!/.bundle/ci-deploy", after: '/.bundle'
+
 # .github/PULL_REQUEST_TEMPLATE.md
 PULL_REQUEST_TEMPLATE_FILE = <<-HEREDOC.strip_heredoc
 Task: [#__TASK_NUMBER__](__ADD_URL_TO_PRODUCTIVE_TASK__)


lovro-bikic (approved):

template.rb
```diff
@@ -480,6 +480,9 @@
 
 append_file '.gitignore', GITIGNORED_FILES
 
+# remove bundler config folder from gitignore
+gsub_file('.gitignore', /# Ignore bundler config\.\n|\/\.bundle\n{2}/, '')
```
d4be4st, line 484: how about we change this to just `.bundle/config` ?
this way your local bundle config will not effect others, but the ci and server ones will still be avaliable.

or we rename the ci and server one to `.bundle-ci/config` and `.bundle-server/config` ?

nikajukic (approved):

template.rb
JureCindro: What about:
```gitignore
/.bundle/*
!/.bundle/ci-build
!/.bundle/ci-deploy
```

JureCindro:
```diff
diff --git a/template.rb b/template.rb
index 57f2f96..77e4802 100644
--- a/template.rb
+++ b/template.rb
@@ -480,8 +480,8 @@ HEREDOC
 
 append_file '.gitignore', GITIGNORED_FILES
 
-# remove bundler config folder from gitignore
-gsub_file('.gitignore', /# Ignore bundler config\.\n|\/\.bundle\n{2}/, '')
+# remove bundler ci config folders from gitignore
+append_to_file '.gitignore', "/*\n!/.bundle/ci-build\n!/.bundle/ci-deploy", after: '/.bundle'
 
 # .github/PULL_REQUEST_TEMPLATE.md
 PULL_REQUEST_TEMPLATE_FILE = <<-HEREDOC.strip_heredoc

PR #98, opened 2022-01-10
Fix bundler app config

d4be4st:
Introducing BUNDLE_APP_CONFIG to deploy broke passenger because it cant find installed gems.

### DIFF
--- a/mina_deploy.rb
+++ b/mina_deploy.rb
@@ -6,7 +6,7 @@ require 'mina/infinum'
 set :application_name, 'awesome_app'
 set :repository, 'git://...'
 set :service_manager, :systemd
-set :bundle_options, ''
+set :bundle_withouts, 'development test ci deploy'
 # set :background_worker, 'dj' / 'sidekiq'
 
 task :staging do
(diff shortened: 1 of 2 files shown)


melcha (approved):

mina_deploy.rb
```diff
@@ -6,7 +6,7 @@
 set :application_name, 'awesome_app'
 set :repository, 'git://...'
 set :service_manager, :systemd
-set :bundle_options, ''
+set 'bundle_withouts', 'development test ci deploy'
```
melcha, line 9: Does it need to be a string, or should we use symbol for consistency?

d4be4st: because of the spaces, yes
until mina iteself changes this to something else (like an array of symobls)

uncoverd (approved):

mina_deploy.rb
lovro-bikic: But Mina [uses](https://github.com/mina-deploy/mina/blob/efc7c9fe53da00e5f9d6c96627faab94a0d63273/tasks/mina/bundler.rb#L7) `:bundle_withouts` key for this option, it's not a string.

d4be4st: oooh, the comment was for the name of the variable, not for the value :D

melcha (approved):

PR #107, opened 2022-01-20
Enable triggering builds manually

nikajukic:
Task: No task

#### Aim
Deploys could be triggered manually from some branch, but builds couldn't.

#### Solution
Add option to trigger a build workflow:

Screenshot 2022-01-20 at 07 43 03

### DIFF
--- a/build.yml
+++ b/build.yml
@@ -1,6 +1,8 @@
 name: Build
 
-on: [push]
+on:
+  workflow_dispatch:
+  push:
 
 jobs:
   build:


build.yml
```diff
@@ -1,6 +1,8 @@
 name: Build
 
-on: [push]
+on:
+  workflow_dispatch:
+  push:
```
cilim, line 5: Not sure if this is the correct syntax, but from looking at the [documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-multiple-events), we should use:

```suggestion
on: [workflow_dispatch, push]
```

cilim (approved):
You're more experienced with GHA at this point so I assume this is some kind of different syntax, but it'll work regardless :)

build.yml
lovro-bikic: Both versions are correct (this was my patch, I just told Nika to open a PR for it). I used this syntax to be consistent with the `deploy-staging` and `deploy-production` workflows.

cilim: Ah, got it! Thanks 👍

PR #108, opened 2022-01-20
Set env variables for generating dox

nikajukic:
Task: No task

#### Aim
Currently generating dox fails when pulling secrets because env variables are not set.
image (13)

#### Solution
Set env variables needed for generating dox during deploy

**This is just a proposal, there's probably a better way to do it?**

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -41,6 +41,14 @@ on:
     secrets:
       SSH_PRIVATE_KEY:
         required: true
+      VAULT_ADDR:
+        required: true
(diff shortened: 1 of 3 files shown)


lovro-bikic (approved):

JureCindro (approved):

nikajukic:
@JureCindro @lovro-bikic 
**UPDATE**
@IKrizic tried this on underline and it didn't work :( we're still getting the same error message

melcha (approved):

nikajukic:
It seems like it should work indeed. Let's try to merge it to v1 and retry again on Ivana's project.

PR #109, opened 2022-01-20
Add a new workflow for generating API docs

nikajukic:
Task: No task

#### Aim
We had a [discussion on Slack](https://infinum.slack.com/archives/C02S24Y5JSV/p1642679196001300) today about generating dox for API apps. 

It was previously done in bin/deploy but now it started failing because of:

- missing env variables for pulling secrets #106 
- missing postgres
- ...

We came to a conclusion to create a new workflow for generating dox.

#### Solution
Reverted #106 because it's not needed anymore since we're adding a new workflow. 

Added a new workflow for publishing docs that [Lovro suggested](https://gist.github.com/lovro-bikic/0f79075f413ef343650d693c27b37f3e) with a few changes:

- removed use_node option because you always need node for generating API docs with dox
- added prepare CI step for pulling secrets
- removed caching rubocop step since we don't need it
- added a new job to deploy-staging.yml workflow (commented out with instructions)

This workflow should be run on successful deploy (defined in deploy-#{env}.yml files with needs: [deploy]).

bin/deploy is also updated and we removed all dox related commands from there
bin/publish_docs now has commands for generating docs

Note: I haven't tried this workflow yet. Please leave a review with suggestions/comments if you have them.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -1,0 +1,96 @@
+name: Publish docs
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  workflow_call:
+    inputs:
+      # Selects the version of Postgres for running tests
+      # See: https://github.com/docker-library/docs/blob/master/postgres/README.md#supported-tags-and-respective-dockerfile-links
+      postgres_image:
+        required: true
+        type: string
+
+      # Sets BUNDLE_APP_CONFIG environment variable
+      # See: https://bundler.io/man/bundle-config.1.html
+      bundle_app_config:
+        required: false
+        type: string
+        default: .bundle/ci-deploy
+
+      # Selects the runner on which the workflow will run
+      # See: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
+      runner:
+        required: false
+        type: string
+        default: ubuntu-20.04
+
+      # Sets the Mina environment (e.g. staging, production)
+      # A task by the same name must exist in config/deploy.rb
+      environment:
+        required: true
+        type: string
+    secrets:
+      SSH_PRIVATE_KEY:
+        required: true
+      VAULT_ADDR:
+        required: true
+      VAULT_AUTH_METHOD:
+        required: true
+      VAULT_AUTH_ROLE_ID:
+        required: true
+      VAULT_AUTH_SECRET_ID:
+        required: true
+
+jobs:
+  publish_docs:
+    name: 'Publish docs'
+    runs-on: ${{ inputs.runner }}
+    env:
+      BUNDLE_APP_CONFIG: ${{ inputs.bundle_app_config }}
+    services:
+      postgres:
+        image: postgres:${{ inputs.postgres_image }}
+        env:
+          POSTGRES_HOST_AUTH_METHOD: trust
+        ports:
+          - 5432:5432
+        options: --name=postgres
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+      - name: Set up Node
+        uses: actions/setup-node@v2
+        with:
+          node-version-file: '.node-version'
+      - name: Prepare node_modules cache
+        uses: actions/cache@v2
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-modules-
+      - uses: webfactory/ssh-agent@v0.5.4
+        with:
+          ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
+      - name: Install JS packages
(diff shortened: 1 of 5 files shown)


.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,97 @@
...
+      postgres_image:
+        required: true
+        type: string
+
+      # Sets BUNDLE_APP_CONFIG environment variable
+      # See: https://bundler.io/man/bundle-config.1.html
+      bundle_app_config:
+        required: false
+        type: string
+        default: .bundle/ci-build
```
nikajukic, line 21: Should we change this to .bundle/ci-deploy-docs where only some of the gems groups are installed?

What do we even need? `test` group + dox?

publish-docs-staging.yml
```diff
@@ -0,0 +1,19 @@
+name: Publish docs staging
```
lovro-bikic, line 1:
```suggestion
name: Publish docs
```

This flow can be used to publish docs on multiple environments (by changing `on.workflow_run.workflows`).

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,97 @@
...
+        required: true
+      VAULT_AUTH_METHOD:
+        required: true
+      VAULT_AUTH_ROLE_ID:
+        required: true
+      VAULT_AUTH_SECRET_ID:
+        required: true
+
+jobs:
+  publish_docs:
```
lovro-bikic, line 48: You have to check that the workflow which called this workflow (deploy) was successful:
```yml
    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
otherwise this will be invoked even when a deploy fails.

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,97 @@
...
+    name: 'Publish docs'
+    runs-on: ${{ inputs.runner }}
+    env:
+      BUNDLE_APP_CONFIG: ${{ inputs.bundle_app_config }}
+      RUBOCOP_CACHE_ROOT: .rubocop-cache
```
lovro-bikic, line 53: We don't need this.

publish-docs-staging.yml
```diff
@@ -0,0 +1,19 @@
...
+  workflow_run:
+    workflows: [Deploy staging]
+    types: [completed]
+
+jobs:
+  build:
+    name: Publish docs
+    uses: infinum/default_rails_template/.github/workflows/publish-docs.yml@v1
+    with:
+      postgres_image: '13.2'
```
lovro-bikic, line 13: `environment` variable is required but it's missing here.

publish-docs-staging.yml
lovro-bikic, line 1: I think it's better that we add this as job to `deploy-staging`, `deploy-uat`, etc., and not as a separate workflow.

Consider this:
```yml
# deploy-staging.yml
jobs:
  deploy:
    name: Deploy
    uses: infinum/default_rails_template/.github/workflows/deploy.yml@v1
    with:
      environment: staging
      branch: staging
      deployers: 'DEPLOY USERS GO HERE' # Example: '@github_username1 @github_username2'
    secrets:
      SSH_PRIVATE_KEY: ${{ secrets.SSH_PRIVATE_KEY_STAGING }}
  publish_docs:
    name: Publish docs
    needs: [deploy]
    uses: infinum/default_rails_template/.github/workflows/publish-docs.yml@v1
    with:
      postgres_image: '13.2'
      environment: staging
    secrets:
      SSH_PRIVATE_KEY: ${{ secrets.SSH_PRIVATE_KEY_STAGING }}
      VAULT_ADDR: ${{ secrets.VAULT_ADDR }}
      VAULT_AUTH_METHOD: ${{ secrets.VAULT_AUTH_METHOD }}
      VAULT_AUTH_ROLE_ID: ${{ secrets.VAULT_AUTH_ROLE_ID }}
      VAULT_AUTH_SECRET_ID: ${{ secrets.VAULT_AUTH_SECRET_ID }}

.github/workflows/publish-docs.yml
lovro-bikic: Note: if we implement [this](https://github.com/infinum/default_rails_template/pull/109/files#r789055512), the we don't need this check.

publish-docs-staging.yml
nikajukic: Aaaa nice, I like this, makes more sense 👌 Will update it.

.github/workflows/publish-docs.yml
nikajukic: Yup, removed it in latest commit

melcha (approved):

cilim (approved):
FYI, we're using postgres 11.12 on Underline staging 🙂

nikajukic:
> FYI, we're using postgres 11.12 on Underline staging 🙂

Yup, I know. This is just a default for new apps :)

cilim:
> > FYI, we're using postgres 11.12 on Underline staging 🙂
> 
> Yup, I know. This is just a default for new apps :)

Just giving you a heads up cause you mentioned you'll test it on Underline 😄 saved you 30 seconds!

.github/workflows/publish-docs.yml
nikajukic: Changed it to .bundle/ci-deploy because we need mina & dox.

PR #113, opened 2022-02-05
Add support for additional environment variables in workflows

lovro-bikic:
#### Aim

On some projects, developers need to pass additional environment variables to a workflow (e.g. `CODECOV_TOKEN`). Currently, this cannot be achieved nicely with GHA because there is no API for passing arbitrary environment variables to a reusable workflow. The only way to pass a variable is to have it defined in the reusable workflow beforehand.

This PR provides a solution which circumvents that limitation.

#### Solution

Add an optional secret `ADDITIONAL_VARIABLES` through which the developers can supply any environment variables they need. The expected structure of the secret is a JSON object, for example:
```json
{ "FOO": "BAR", "BAZ": "${{ secrets.BAZ }}" }
```
which will be parsed by the reusable workflow using [jq](https://stedolan.github.io/jq/). Keys of the object will become environment variable names, and the values of the object will become environment variable values. In the above example, that means the following env variables would be exported: `FOO=BAR` and `BAZ=THE_SECRET_VALUE_OF_BAZ`.

The following examples show how the new secret works.

1. The secret hasn't been passed (the step is skipped):
Screenshot 2022-02-05 at 10 47 35

2. The secret is invalid JSON (the workflow aborts, an error message is output):
Screenshot 2022-02-05 at 10 39 17

3. The secret is valid JSON (environment variables are set, and a success message is printed for each variable):
Screenshot 2022-02-05 at 10 40 48

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,6 +49,12 @@ on:
       VAULT_AUTH_SECRET_ID:
         required: true
 
+      # Additional environment variables set in the workflow
+      # Format: JSON object with string values (key becomes env variable name, value becomes env variable value)
+      # Example: '{ "FOO": "BAR", "BAZ": "${{ secrets.BAZ }}" }' 
+      ADDITIONAL_VARIABLES:
+        required: false
+
 jobs:
   build:
     name: 'Build'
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -65,6 +71,18 @@ jobs:
...
+          ADDITIONAL_VARIABLES: ${{ secrets.ADDITIONAL_VARIABLES }}
+        if: ${{ env.ADDITIONAL_VARIABLES }}
+        run: >
+          if echo '${{ env.ADDITIONAL_VARIABLES }}' | jq >/dev/null 2>&1; then
+            echo '${{ env.ADDITIONAL_VARIABLES }}' | jq -r 'to_entries[] | "\(.key) \(.value)"' | \
+              while read -r key value; do echo "$key=$value" >> $GITHUB_ENV && echo "Variable $key has been set"; done
+          else
+            echo "ADDITIONAL_VARIABLES secret you supplied is not a valid JSON object. Check the formatting of the secret."
+            exit 1
+          fi
```
lovro-bikic, line 85: How this works:
1. Check if the supplied secret is valid JSON (line 79):
  a. If not, abort with an error message (lines 83-84)
  b. If yes, then:
    1. Transform the object into key-value pairs (line 80)
    2. For each pair, [set an environment variable](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable) and print a success message (line 81)

nikajukic (approved):

vr4b4c (approved):

cilim (approved):

PR #115, opened 2022-02-09
Add Slack notifications support to reusable workflows

lovro-bikic:
Task: [#510](https://app.productive.io/1-infinum/tasks/2224963)

#### Aim

Allow reusable workflows to send Slack notifications upon completion of workflow runs.

#### Solution

Add a step to build and deploy reusable workflows which invokes the Slack [postMessage](https://api.slack.com/methods/chat.postMessage) API endpoint to send a message with the workflow run result. The channel to which the notification will be sent can be configured with an input. You can also configure what kind of notifications you want to receive: just successful runs, just failed runs, or both.

Build notifications:
Screenshot 2022-02-09 at 09 30 28
Screenshot 2022-02-09 at 09 30 55

Deploy notifications:
Screenshot 2022-02-09 at 09 30 40
Screenshot 2022-02-09 at 09 30 33


vr4b4c (approved):
🤘

nikajukic (approved):
Should we make Slack notifications mandatory? (basically make slack_notification_channel a mandatory param)

I think all projects should have a channel where we see the notifications, if they don't have it, we can fallback to #backend-apps-notifications

lovro-bikic:
I wouldn't make it mandatory. On the project that I work on only I receive notifications and it's much easier for me to just look at the GH page. Slack notifications would be too much noise for me.

PR #116, opened 2022-02-21
Add node toggle to publish docs github action

JureCindro:
Task: NA

#### Aim
I suggest that we do not use the full package.json install to build the documentation if the application does not need any yarn dependencies for the dox specs to pass. The package.json should contain runtime depencencies and development dependencies, and arguably the aglio or redoc-cli used to build the documentation fits in neither of those groups.

#### Solution
Added a use_node input that defaults to true, similar to how it is done in the build workflow. This enables us to use node for runtime dependencies needed by the dox specs to successfully run. 
Added global install of redoc-cli to the publish_docs script. This way the publish_docs github action does not need to install all development dependencies, making it faster.

The changes are backwards compatible.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -27,6 +27,12 @@ on:
         type: string
         default: ubuntu-20.04
 
+      # Determines whether to install Node and run `yarn install`
+      use_node:
+        required: false
+        type: boolean
+        default: true
+
       # Sets the Mina environment (e.g. staging, production)
       # A task by the same name must exist in config/deploy.rb
       environment:
@@ -93,20 +99,23 @@ jobs:
           bundler-cache: true
       - name: Set up Node
         uses: actions/setup-node@v2
+        if: ${{ inputs.use_node }}
         with:
           node-version-file: '.node-version'
       - name: Prepare node_modules cache
         uses: actions/cache@v2
+        if: ${{ inputs.use_node }}
         with:
           path: node_modules
           key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
           restore-keys: |
             ${{ runner.os }}-modules-
+      - name: Install JS packages
+        if: ${{ inputs.use_node }}
+        run: yarn install --frozen-lockfile
       - uses: webfactory/ssh-agent@v0.5.4
         with:
           ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
-      - name: Install JS packages
-        run: yarn install --frozen-lockfile
       - name: Prepare CI
         run: bin/prepare_ci
         if: hashFiles('bin/prepare_ci') != ''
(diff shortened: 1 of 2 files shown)


template.rb
```diff
@@ -225,19 +225,15 @@
   echo "=========== setting env variables ==========="
   environment=$1
 
-  #############################################
-  # Uncomment this if you need to publish dox #
-  # Delete not needed environment             #
-  #############################################
-  #
-  # if [[ $environment =~ ^(production|uat|staging)$ ]]; then
```
JureCindro, line 233: Removed the conditional since the github action has a run condition already, so this seems redundant.

template.rb
```diff
@@ -225,19 +225,15 @@
...
-  #   echo "=========== rails db:test:prepare ==========="
-  #   time bundle exec rails db:test:prepare
-  #
-  #   echo "=========== mina dox publish ==========="
-  #   time bundle exec mina $environment ssh_keyscan_domain
-  #   time bundle exec mina $environment dox:publish
-  # fi
+  echo "=========== install redoc ==========="
+  time yarn global add redoc-cli
```
lovro-bikic, line 229: `redoc-cli` should be a dependency in `package.json` for two reasons:
1. you already run `yarn install` so `package.json` dependencies get installed anyway
2. we want to control the version of `redoc-cli`. This command will always install the latest version of `redoc-cli` which might introduce breaking changes. We want our CI to be reproducible, so the versio needs to be fixed.

.github/workflows/publish-docs.yml
```diff
@@ -27,6 +27,11 @@ on:
         type: string
         default: ubuntu-20.04
 
+      use_node:
```
lovro-bikic, line 30: Please add a short comment explaining the variable. We have this in other workflows:
```yaml
# Determines whether to install Node and run `yarn install`
use_node:
```

template.rb
lovro-bikic: > A common scenario is that we have an API + a custom admin dashboard. API is documentated, admin dashboard needs some js libraries. Does it make sense to load all those libraries for API documentation if the documentation spec do not need it? Doesn't it just make specs slower?

The library isn't loaded just because it's specified in package.json. It's loaded only when you require it in your JS files. So until you add `import 'redoc-cli'` somewhere in your JS, it won't affect your specs. 

Performance for fetching the library is also not an issue since we cache node modules.

> yarn global add redoc-cli@0.13.7

This makes the action reproducible, but you'll most likely forget to update the dependency from time to time. If it's in package.json, Dependabot will remind you (and you can use `yarn outdated`).

JureCindro: Makes sense. Good point on the cache, so only the first run is slower 👌. Ok so I propose that we still keep the `use_node` toggle for compatibility reasons for projects that might not have a `package.json`, and remove the manual installation from the template `publish-docs`.

lovro-bikic (approved):

PR #119, opened 2022-04-06
Extend the build image to allow rails env overrides

JureCindro:
Task: NA

#### Aim
We might require different naming for the ecs environment from the environment we're using in rails and from the branch we're deploying from. 

#### Solution
Added rails_env option.
Also changed the workflow a bit so that we can use it only to test the
main branches without actual deploys.


uncoverd (approved):
This is fine.

PR #120, opened 2022-04-06
Fix the deploy image cache from test image

JureCindro:
Task: NA

#### Aim
We "finalized" the test docker image before building the development image. We forgot to fix the cache-from declaration to take that into account.


uncoverd (commented):
🙆‍♂️

uncoverd (approved):

PR #125, opened 2022-07-01
[dOKR #582] Add strong_migrations

djordjelacmanovic:
Task: [#582](https://app.productive.io/1-infinum/tasks/2752427)

#### Aim

As part of the containerization discussion we came to the conclusion that running safe migrations in deployments is important for all applications and therefore `strong_migrations` should be a part of the template.

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -240,6 +240,7 @@ append_to_file 'Gemfile', after: /gem "rails".*\n/ do
     gem 'pry-byebug'
     gem 'pry-rails'
     gem 'secrets_cli', require: false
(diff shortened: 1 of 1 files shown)


djordjelacmanovic:
> Part of strong_migrations setup is running `rails generate strong_migrations:install`.
> 
> 
> 
> We should add that to the bottom of the template.rb as well.
> 
> 
> 
> 

Good catch, it 100% slipped my mind 🙈

uncoverd (approved):
👌

PR #128, opened 2022-07-27
[dOKR #595] Add ECS connection script

djordjelacmanovic:
Task: [#595](https://app.productive.io/1-infinum/tasks/2752442)

#### Aim

I've copied over the ECS connection script from [example app](https://github.com/infinum/rails-jsonapi-example-app/pull/87) and made it more generic w/ instructions on how to get it working.


docker/connect_to_container
```diff
@@ -0,0 +1,22 @@
...
+
+# follow the instructions in Handbook guide
+# https://infinum.com/handbook/rails/containerization/deployment/connecting-to-a-container
+# then replace the placeholder variables below
+
+export AWS_PROFILE=_PROJECT_SPECIFIC_PROFILE_
+
+CLUSTER_NAME=${1:-_DEFAULT_CLUSTER_NAME_}
+SERVICE_NAME=_YOUR_ECS_SERVICE_NAME_
+CONTAINER_NAME=_CONTAINER_NAME_IN_TASK_
```
d4be4st, line 14: in our coinmara experice we had:
one cluster
multiple services
each service had one container

so the cluster_name would not change
service_name would be equal to container_name

so how about this?

```suggestion
CLUSTER_NAME=${2:-_DEFAULT_CLUSTER_NAME_}
SERVICE_NAME=${1:-_DEFAULT_CLUSTER_NAME_}
CONTAINER_NAME=${3:-$SERVICE_NAME}
```

docker/connect_to_container
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+# usage: bin/connect_to_container - connect to default specified cluster
+#        bin/connect_to_container cluster_name - connect to specific cluster
+
+# follow the instructions in Handbook guide
+# https://infinum.com/handbook/rails/containerization/deployment/connecting-to-a-container
+# then replace the placeholder variables below
+
+export AWS_PROFILE=_PROJECT_SPECIFIC_PROFILE_
```
d4be4st, line 10: AWS_PROFILE is already used widely for other stuff as well, would probably remove this line.

docker/connect_to_container
```diff
@@ -0,0 +1,22 @@
...
+
+CLUSTER_NAME=${1:-_DEFAULT_CLUSTER_NAME_}
+SERVICE_NAME=_YOUR_ECS_SERVICE_NAME_
+CONTAINER_NAME=_CONTAINER_NAME_IN_TASK_
+
+task_arn=$(aws ecs list-tasks --cluster $CLUSTER_NAME --service-name $SERVICE_NAME | jq -r '.taskArns[0]')
+aws ecs execute-command  \
```
d4be4st, line 17: would wrap this in:
```
if [[ -z "$task_arn" ]]
  then print "No tasks found"
else
  print "Connecting to task $task_arn ..."
  aws ecs ...
fi
```

d4be4st (changes requested):

docker/connect_to_container
djordjelacmanovic: Good point, though I made it exit with a non-zero code if no tasks

d4be4st: I know, but some devs do not have fancy prompts with colors :)

d4be4st (approved):

PR #129, opened 2022-07-29
[dOKR #591] Remove the need for Vault

djordjelacmanovic:
Task: [#591](https://app.productive.io/1-infinum/tasks/2752437)

#### Aim

During containerization efforts we have concluded that we will not be using Vault, so we want to make it an opt-in tool in the template.

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -18,8 +18,6 @@ README_MD = <<-HEREDOC.strip_heredoc
   * Yarn (defined in [package.json](https://classic.yarnpkg.com/en/docs/package-json/#toc-engines))
 
 ## Setup
-Before:
-  * ensure you have read permissions for vault
 
 Run:
 ```bash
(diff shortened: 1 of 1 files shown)


template.rb
```diff
@@ -425,8 +382,6 @@
   dump.rdb
   logfile
   .DS_Store
-  # Ignore application configuration
-  config/application*.yml
```
d4be4st, line 429: We still need to ignore config/application.yml files.

1. You might push a credential while you are locally developing.
2. it will override environment variables on ECS.

My suggestion is to use a sample copying.
have `config/application.example.yml` which has all the required keys but values are empty.
in `bin/setup` do `mv config/application.example.yml config/application.yml`

we will also need this file for CI while testing

d4be4st (changes requested):

uncoverd:
@djordjelacmanovic what are our next steps here, do you need more reviews or do you just need to address Štef's comments?

djordjelacmanovic:
> @djordjelacmanovic what are our next steps here, do you need more reviews or do you just need to address Štef's comments?

That's it, unless anyone has additional comments

d4be4st (approved):

PR #132, opened 2023-01-19
Add libvips install

mislavk:
Task: No task

#### Aim

- Add `libvips` to build container to support tests that require `libvips`

#### Solution

- Install libvips42 in the build container if uses_libvips variable set

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -120,12 +120,13 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-rubocop-cache-
       - name: Fetch Aptfile
+        if: ${{ startsWith(inputs.runner, 'ubuntu' && inputs.aptfile }}
         id: apt_packages
         uses: juliangruber/read-file-action@v1
         with:
           path: ${{ inputs.aptfile }}
       - name: Install dependencies
-        if: ${{ startsWith(inputs.runner, 'ubuntu') &&  inputs.aptfile }}
+        if: ${{ startsWith(inputs.runner, 'ubuntu') && inputs.aptfile }}
         run: sudo apt install $aptfile
         env:
           aptfile: "${{ steps.apt_packages.outputs.content }}"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,12 @@ on:
         type: boolean
         default: true
 
+      # Determines whether to install libvips42 and tools
+      use_libvips:
+        required: false
+        type: boolean
+        default: false
+
       # Sets BUNDLE_APP_CONFIG environment variable
       # See: https://bundler.io/man/bundle-config.1.html
       bundle_app_config:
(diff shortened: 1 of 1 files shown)


cilim (approved):

djordjelacmanovic (approved):
~Looks good. Dockerized apps install deps from an appropriate `Aptfile` so this is a welcome addition for the rest.~

djordjelacmanovic:
Come to think of it, putting this in an Aptfile might be a useful feature for all apps. We do have a directory `.github` within projects to define the GH configuration incl. the workflows. 

Would it make sense to put an `Aptfile` there and install dependencies from that? I'm thinking of cases where additional dependencies are needed e.g. `libyaml` that was removed from [psych 5.0.0](https://github.com/ruby/psych/releases/tag/v5.0.0) and has to be installed beforehand.

Thoughts @mislavk @nikajukic @cilim @vr4b4c @uncoverd ?

djordjelacmanovic (changes requested):
IMHO this should be parametrized in some way so it can be used by projects that won't just have `libvips` dependency. https://github.com/infinum/default_rails_template/pull/132#issuecomment-1398244253

vr4b4c:
Agree with @djordjelacmanovic , we should look toward adding the ability to specify general system dependencies. In practice, not every new system dependency (or group of dependencies) should require us to add a new "workflow input"

uncoverd:
Agreed 👍 The template could have a default Aptfile and projects could include their own (e.g. CustomAptfile) and set just one flag if they want to do anything custom

mislavk:
We could use something like [read-file-action](https://github.com/juliangruber/read-file-action) to read the output of the aptfile (or just `run: cat FILE`) if the file exists and install all the packages. I'm not sure what the estimate would be, I'd say 3-4 hours with testing the pipeline.

Which packages would be in the default list? Here's [a list](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#installed-apt-packages) of included packages with ubuntu-latest runner.

mislavk:
@djordjelacmanovic @cilim @uncoverd @vr4b4c @nikajukic 
Where do you want to store that Aptfile? I would recommend `.github/Aptfile`. We can also set this as a variable and use `aptfile = ".github/Aptfile"` instead of `use_libvips`. I've prepared all usecases.

mislavk:
I've added the Aptfile support, when the variable is defined and the runner is an ubuntu container, it will install all the packages in the Aptfile on the runner.

Here's an example of the build:
```yaml
jobs:
  build:
    name: Build
    uses: infinum/default_rails_template/.github/workflows/build.yml@v1
    with:
      aptfile: ".github/Aptfile"
```
Aptfile:
```
libvips42
libvips-tools
```

cilim (approved):
Looking good 🙂 

I'll test this out on Underline next week if all other reviewers agree this is good to go.

djordjelacmanovic (approved):
LGTM

vr4b4c (approved):
🤘

PR #133, opened 2023-01-20
Update delete cache workflow timeout and concurrency

lovro-bikic:
No task

#### Problem

If you merge multiple PRs to master quickly, multiple delete-cache jobs will start, but all except the last one will be cancelled because of concurrency settings. Because of this some cache entries won't be deleted.

image

Also, timeout hasn't been configured for this workflow.

#### Solution

Remove concurrency settings so that all jobs run.

Add a 2-minute timeout to the workflow.

### DIFF
--- a/.github/workflows/delete-cache.yml
+++ b/.github/workflows/delete-cache.yml
@@ -1,9 +1,5 @@
 name: Delete Cache
 
-concurrency:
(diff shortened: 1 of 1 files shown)


nikajukic (approved):

lovro-bikic:
Released on v1.

PR #138, opened 2023-06-02
Task #876 - Add license_finder to template

anamarijabalaban:
Task: [#876](https://latest.productive.io/1-infinum/time/me/task/4779901?tour=time-week-layout)

#### Aim
Add license_finder to the template.

#### Solution
Added license_finder gem and prepared a list of restricted and permitted licenses. Also, license_finder is added to pre-commit hooks.

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -726,9 +726,11 @@ run 'overcommit --sign pre-push'
 run 'bundle exec rubocop -A'
 
 # add a list of permitted licenses to license_finder configuration file
-run 'bundle exec license_finder permitted_licenses add "New BSD" "Simplified BSD" ' \
-    'ruby "2-clause BSDL" "Python-2.0" MIT "Apache 2.0" ISC CC0-1.0 CC-BY-3.0 ' \
-    '"Mozilla Public License 2.0" LGPL'
+run 'bundle exec license_finder permitted_licenses add "New BSD" "Simplified BSD" ruby "2-clause BSDL" ' \
+    '"Python-2.0" MIT "Apache 2.0" ISC CC0-1.0 CC-BY-3.0 "Mozilla Public License 2.0" LGPL MIT-0 ' \
+    '"Brakeman Public Use License" Apache "(CC-BY-4.0 AND OFL-1.1 AND MIT)" CC-BY-4.0 ' \
+    '"MPL-2.0 OR Apache-2.0" "(MIT AND Zlib)" "(WTFPL OR MIT)" "PayPal SDK License" ' \
+    '"MIT (http://mootools.net/license.txt)" "(MIT AND CC-BY-3.0)" 0BSD BSD '
 
 # add a list of restricted licenses to license_finder configuration file
 run 'bundle exec license_finder restrict add GPL-2.0 GPL-3.0'
--- a/template.rb
+++ b/template.rb
@@ -266,6 +266,7 @@ append_to_file 'Gemfile' do
       gem 'brakeman', require: false
       gem 'bundler-audit', require: false
       gem 'rubocop-infinum', require: false
+      gem 'license_finder', require: false
     end
   HEREDOC
 end
@@ -605,6 +606,9 @@ if yes?('Will this application have a frontend? [No]', :green)
 
   create_file '.node-version', node_version
 
+  # add infinum gems to approved gems list because currently their license is UNKNOWN
+  run 'bundle exec license_finder approvals add @infinumrails/stylelint-config-scss @infinumrails/eslint-config-js'
+
   PACKAGE_JSON_FILE = <<~HEREDOC
     {
       "name": "#{app_name}",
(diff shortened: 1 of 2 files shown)


template.rb
```diff
@@ -718,3 +727,19 @@ def run
...
 # Fix default rubocop errors
 run 'bundle exec rubocop -A'
+
+# add development, test and ci gem groups to ignored_groups when checking license types
+run 'bundle exec license_finder ignored_groups add development'
+run 'bundle exec license_finder ignored_groups add test'
+run 'bundle exec license_finder ignored_groups add ci'
+
+# add a list of permitted licenses to license_finder configuration file
+run 'bundle exec license_finder permitted_licenses add "New BSD" "Simplified BSD" ' \
```
vr4b4c, line 737: Where did all these licenses come from? How did you decide which ones are permitted?

template.rb
```diff
@@ -718,3 +727,19 @@ def run
...
+    'ruby "2-clause BSDL" "Python-2.0" MIT "Apache 2.0" ISC CC0-1.0 CC-BY-3.0 ' \
+    '"Mozilla Public License 2.0" LGPL'
+
+# add a list of restricted licenses to license_finder configuration file
+run 'bundle exec license_finder restrict add GPL-2.0 GPL-3.0'
+
+# add infinum gems to approved gems list because currently their license is UNKNOWN
+run 'bundle exec license_finder approvals add enumerations'
```
vr4b4c, line 745: I wouldn't add this by default, only if the project actually depends on `enumerations`

vr4b4c:
Should we also run a license check in CI/CD pipeline?

template.rb
```diff
@@ -605,6 +606,9 @@ def run
 
   create_file '.node-version', node_version
 
+  # add infinum gems to approved gems list because currently their license is UNKNOWN
+  run 'bundle exec license_finder approvals add @infinumrails/stylelint-config-scss @infinumrails/eslint-config-js'
```
uncoverd, line 610: I think this could be added as a post install warning (e.g. "you might need to add config-scss and eslint-config-js to the approved list manually with the following command"), I'm not sure it makes sense to always add it

template.rb
```diff
@@ -647,6 +651,11 @@ def run
     on_warn: fail
     required_executable: 'node_modules/.bin/stylelint'
     command: ['node_modules/.bin/stylelint']
+
+  LicenseFinder:
+    enabled: true
+    on_warn: fail
+    command: ['bundle', 'exec', 'license_finder']
```
uncoverd, line 658: How long does the command usually take on a bigger or smaller project?

template.rb
```diff
@@ -718,3 +727,19 @@ def run
 
 # Fix default rubocop errors
 run 'bundle exec rubocop -A'
+
+# add development, test and ci gem groups to ignored_groups when checking license types
+run 'bundle exec license_finder ignored_groups add development'
+run 'bundle exec license_finder ignored_groups add test'
+run 'bundle exec license_finder ignored_groups add ci'
```
uncoverd, line 734: I think an audit would flag license misuse for these dependencies as well. How come you decided to ignore them?

template.rb
uncoverd: I think we have a pending OKR to go over the default list with Legal. @vlakre is this your list?

uncoverd (changes requested):

template.rb
vlakre: @uncoverd yep, I've compiled this list when adding this to Atlas. I went over the licenses from the gems we had, checked each one, and added it to the list. There might be other licenses that are ok to use but are not covered here.
If there's a chance to check these with Legal, by all means 💪

anamarijabalaban: Yes, that is the list of permitted licenses that Kreso found on Atlas, but I think that I will need to expand this list because in the meantime I find some other non-restrictive licenses on some gems that are not present on Atlas. (on Loop and NLBL)

Also, I am planning to sit with our legal and check with them which group of licensees we need to avoid.

template.rb
anamarijabalaban: I wasn't 100% sure about this, so I talked with Kreso about this (since he was adding `lincense_finder` on Atlas and researching this topic before). In the end, I decided to ignore development, test and ci gems because they are not directly used in published app (staging and production), but I will definitely check this with our legal.

Also, an interesting thing is that `bundler-audit ` gem has a GPL-3.0 license which is a very strong copyleft license.

template.rb
anamarijabalaban: I measure that locally on some projects that I have access:
loop - 23 sec
atlas (we have implemented some ignored groups and other rules) - 30 sec
lmt slovenia - 27sec
nlbl - 15 sec (no js dependencies)

vlakre: @uncoverd it will only be run when the Gemfile changes. So the long waiting time will only affect the commit where you add the new gem.

uncoverd:
@anamarijabalaban what are your thoughts on adding this on CI? Historically we used CI as the last line of defense as overcommit is not always installed. Afaik we'd need to add a command to the CI steps and introduce a `bin/license` file to run the command

anamarijabalaban:
@uncoverd I'm just preparing a separate PR that adds a GA workflow file for license scanning on the project that already has configured license_finder gem.

nikajukic:
@vr4b4c @uncoverd please do a final round of review before we merge this.

uncoverd (approved):

vr4b4c (approved):

nikajukic:
@anamarijabalaban these changes will need to be moved to https://github.com/infinum/rails-template, our new template repository

PR #143, opened 2023-09-27
Use ref name instead of SHA for rubocop cache

lovro-bikic:
No task

#### Aim & Solution

Use ref name (branch name) instead of latest branch SHA for rubocop cache key. This will ensure there's only a single rubocop cache entry per branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -111,7 +111,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ${{ env.RUBOCOP_CACHE_ROOT }}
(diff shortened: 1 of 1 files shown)


vr4b4c (commented):
Curious, how does this affect running builds on multiple commits on the same branch?

lovro-bikic:
It doesn't really since multiple commits on the same branch already used the same cache because of the `restore-keys` param, the only difference is that now that cache is stored under a single entry instead of split per commit.

vr4b4c (approved):

PR #145, opened 2023-10-11
Task #905 - Add scan-licenses.yml GA workflow

anamarijabalaban:
Task: [#905](https://app.productive.io/1-infinum/time/me/task/5805311?tour=time-week-layout)

#### Aim
Add GA workflow that will run license scanning in the CI/CD pipeline for projects that already have configured license_finder gem.

#### Solution
Added scan-licenses.yml GA workflow that will be triggered when the Gemfile or package.json file has been changed.

First I tried to run license_finder only by using this command: `bundle exec license_finder` but I got some errors with missing nokogiri so I found this solution that includes adding x86_64-linux platform to the bundler. [License finder issue with this solution](https://github.com/pivotal/LicenseFinder/issues/828).
Screenshot 2023-10-11 at 08 58 27

### DIFF
--- a/.github/workflows/scan-licenses.yml
+++ b/.github/workflows/scan-licenses.yml
@@ -1,10 +1,15 @@
 name: Scan licenses
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 on:
   workflow_call:
 
 jobs:
   scanning:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     name: License scanning
     steps:
       - name: Git checkout
--- a/.github/workflows/scan-licenses.yml
+++ b/.github/workflows/scan-licenses.yml
@@ -1,0 +1,21 @@
+name: Scan licenses
+on:
+  workflow_call:
+
+jobs:
+  scanning:
+    runs-on: ubuntu-20.04
+    name: License scanning
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v3
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+      - name: Run license scanning
+        run: |
+          bundle config --local deployment false
+          bundle lock --add-platform x86_64-linux
+          bundle config --local deployment true
+          bundle exec license_finder
--- a/scan-licenses.yml
+++ b/scan-licenses.yml
@@ -1,11 +1,12 @@
 name: Scan licenses
 
 on:
+  workflow_dispatch:
   push:
     paths:
       - 'Gemfile*'
       - 'package.json'
 
 jobs:
-  cleanup:
+  scanning:
     uses: infinum/default_rails_template/.github/workflows/scan-licenses.yml@v1
--- a/scan-licenses.yml
+++ b/scan-licenses.yml
@@ -1,0 +1,11 @@
+name: Scan licenses
+
+on:
+  push:
+    paths:
+      - 'Gemfile*'
+      - 'package.json'
+
+jobs:
+  cleanup:
+    uses: infinum/default_rails_template/.github/workflows/scan-licenses.yml@v1
--- a/template.rb
+++ b/template.rb
@@ -493,6 +493,7 @@ end
 
 get("#{BASE_URL}/build.yml", '.github/workflows/build.yml')
 get("#{BASE_URL}/delete-cache.yml", '.github/workflows/delete-cache.yml')
+get("#{BASE_URL}/scan-licenses.yml", '.github/workflows/scan-licenses.yml')
 
 ## Docker
 if no?('Will this application use Docker? [Yes]', :green)


anamarijabalaban:
Later, I figured out that this would be more useful if a workflow could be run on any project no matter if it has configured license_finder or not, so maybe I could prepare a workflow that will install, configure license_finder gem and run license scanning. What do you think?

.github/workflows/scan-licenses.yml
```diff
@@ -0,0 +1,21 @@
...
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+      - name: Run license scanning
+        run: |
+          bundle config --local deployment false
+          bundle lock --add-platform x86_64-linux
+          bundle config --local deployment true
+          bundle exec license_finder
```
lovro-bikic, line 21: I'd actually like to have this as a script in the project, similar to e.g. `bin/lint`: https://github.com/infinum/default_rails_template/blob/ce3d22884a1e32c9fcf0ad93ddc8db1ed5939ca6/template.rb#L164-L178

That way developers can run the script locally and modify it if needed.

scan-licenses.yml
```diff
@@ -0,0 +1,11 @@
+name: Scan licenses
+
+on:
+  push:
+    paths:
+      - 'Gemfile*'
+      - 'package.json'
+
+jobs:
+  cleanup:
```
lovro-bikic, line 10:
```suggestion
  scanning:
```

.github/workflows/scan-licenses.yml
```diff
@@ -0,0 +1,21 @@
+name: Scan licenses
+on:
+  workflow_call:
+
+jobs:
+  scanning:
+    runs-on: ubuntu-20.04
```
lovro-bikic, line 7: Can we try `ubuntu-latest` here?

.github/workflows/scan-licenses.yml
```diff
@@ -0,0 +1,21 @@
+name: Scan licenses
+on:
```
lovro-bikic, line 2: Let's keep it tidy:
```suggestion
name: Scan licenses

on:
```

and let's also add `concurrency` setup: https://github.com/infinum/default_rails_template/blob/ce3d22884a1e32c9fcf0ad93ddc8db1ed5939ca6/.github/workflows/publish-docs.yml#L3-L5

scan-licenses.yml
```diff
@@ -0,0 +1,11 @@
+name: Scan licenses
+
+on:
+  push:
```
lovro-bikic, line 4: I think `workflow_dispatch` event would also be useful here, so developers can trigger it manually as well.

.github/workflows/scan-licenses.yml
lovro-bikic, line 1: How are the developers notified that there are issues with some licenses?

nikajukic: Agree, we should setup a Slack notification here. 

DevOps used https://github.com/rtCamp/action-slack-notify

nikajukic:
@uncoverd @vr4b4c please leave your review

uncoverd:
@lovro-bikic 's suggestions are great, I think once we resolve those we're good to push this 👍 

As far as the Slack notifications are concerned, we'd get a failing build notification and a visual notification on the PR. On top of that we have after commit hooks defined in the other PR so I don't think another Slack notification is required

.github/workflows/scan-licenses.yml
vr4b4c: I don't think Slack notification is necessary for this step.

.github/workflows/scan-licenses.yml
```diff
@@ -0,0 +1,21 @@
...
+        uses: actions/checkout@v3
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+      - name: Run license scanning
+        run: |
+          bundle config --local deployment false
+          bundle lock --add-platform x86_64-linux
+          bundle config --local deployment true
```
vr4b4c, line 20: This block looks like it's trying to perform a trick/hack. Why do we need this? Is there a more conventional way of achieving the same result?

.github/workflows/scan-licenses.yml
vr4b4c: For simple one/two-liners, I find it's more complex when CI workflow is referencing project scripts. GH workflow is coupled with the default_rails_template, and it makes it less portable.

In this case, I'm OK with both approaches, but in general, I'm in favor of workflow being self-sustainable, and project scripts should be used only where necessary.

vr4b4c (commented):
Did we consider running this check as part of the build workflow?

.github/workflows/scan-licenses.yml
anamarijabalaban: First, I tried just to call `bundle exec license_finder` but I got some missing nokogiri error https://github.com/infinum/default_rails_template/pull/145#issue-1937937171

.github/workflows/scan-licenses.yml
anamarijabalaban: @nikajukic I tried to use https://github.com/rtCamp/action-slack-notify for sending slack notifications but it looks like they require us to install this https://slack.com/apps/A0F7XDUAZ-incoming-webhooks?tab=settings&next_id=0

.github/workflows/scan-licenses.yml
vr4b4c: Seems like you uncovered a problem with our setup. This is a native extension compilation problem, which originates from our `template.rb` specifically [here](https://github.com/infinum/default_rails_template/blob/master/template.rb#L702).

Nokogiri [installation docs](https://nokogiri.org/tutorials/installing_nokogiri.html) contains compelling arguments to explicitly list platforms making bundler install precompiled binaries instead of compiling deps natively.

@lovro-bikic your comment [here](https://github.com/infinum/default_rails_template/pull/124#issuecomment-1169875889) was the trigger to move from explicitly listing platforms to using `ruby`. Can you please take a look at this problem and add your thoughts on the topic?

anamarijabalaban: Hey @lovro-bikic can you please check these problems?

anamarijabalaban:
> Did we consider running this check as part of the build workflow?

I think it would be better to have this license scanning as a separate workflow, so that we don't scan licenses on each build. In this way, we can scan licenses only when the Gemfile or package.json file includes some changes.

vr4b4c:
Makes sense. This repository is being replaced with https://github.com/infinum/rails-template, can you please port your PR there?

PR #146, opened 2024-07-10
Feature/launch_rails_app_in_deploy_rb

vlakre:
#### Aim
Update deploy.rb script and force the load of the rails environment to catch missing secrets or anything similar before deploying.

#### Solution
Run random ruby script with rails runner to trigger the rails env load.

### DIFF
--- a/mina_deploy.rb
+++ b/mina_deploy.rb
@@ -36,6 +36,7 @@ task :deploy do
     invoke :'bundle:install'
     command 'yarn install'
     invoke :'secrets:pull'
+    invoke :load_rails_environment
     invoke :'rails:db_migrate'
     invoke :'rails:assets_precompile'
     invoke :'deploy:cleanup'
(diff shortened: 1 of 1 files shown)


uncoverd (approved):
What's the time penalty?

vlakre:
@uncoverd about 10sec, according to the Atlas deployment logs.

vr4b4c (approved):
I understand this solves practical challenge, but you should be aware that this isn't in sync with how application is started in the online env. Here you are starting the app using the `bundle exec` while in the online env it'll be managed using the web application server (probably Passenger). In other words, you could miss certain types of  startup errors (eg. stdlib package version mismatches).

vlakre:
@vr4b4c good point. Is there a way for us to run the app in a way that is in sync with the online env? It would be great to catch package version mismatches and similar issues.

vr4b4c:
I think the problem you are trying to solve should be handled on the infrastructure layer with good observability setup, in this particular case monitoring health-check endpoint + alerting rules to notify the failure

vlakre:
after consulting with @vr4b4c, I see that adding this to the default template is unnecessary as the new projects are moving away from mina. Closing this PR.

PR #149, opened 2024-11-19
[Task #960] Introduce a feature flag system 

Benaaaaa:
Task: [#960](https://app.productive.io/1-infinum/task/7909978)

#### Aim
We need to introduce a feature flag management system to the default Rails template. 

#### Solution
Introduce flipper to the project
Do not set ! on secrets regarding flipper cloud, as they are not necessary for Flipper to be working
Update the readme regarding Flipper

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,16 @@ For build workflow to work, the following secrets must exist (usually set up by 
 
 For deploy workflows, you need to generate private/public SSH key pairs for each environment. Public key should be added to the server to which you're deploying. Private key should be added as a secret to GitHub and named `SSH_PRIVATE_KEY_#{ENVIRONMENT}`, where `ENVIRONMENT` is replaced with an appropriate environment name (`STAGING`, `PRODUCTION`, etc.).
 
+### Flipper
+
+This template uses Flipper for feature flag management. There are two ways for Flipper to function: **Cloud** and **Cloudless**. Cloudless will work out of the box, but if we want Flipper to be communicating with Flipper cloud there are a couple of secrets that need to be set:
+
+- `FLIPPER_CLOUD_TOKEN` - needed for our App to be able to communicate with Flipper Cloud
+- `FLIPPER_CLOUD_SYNC_SECRET` - used if we want Flipper Cloud to send a webhook request to our app telling it to sync with Cloud
+
+Both of these secrets are available inside the Flipper Cloud dashboard, and need to be configured per environment
+
+
 #### Slack notifications
 
 Build and deploy workflows can send Slack notifications upon completion of workflow runs. To enable this, add the following to the workflow:
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ For deploy workflows, you need to generate private/public SSH key pairs for each
 
 ### Flipper
 
-This template uses Flipper for feature flag management. There are two ways for Flipper to function: **Cloud** and **Cloudless**. Cloudless will work out of the box, but if we want Flipper to be communicating with Flipper cloud there are a couple of secrets that need to be set:
+This template uses Flipper for feature flag management. There are two ways for Flipper to function: **Cloud** and **Cloudless(Self-hosted)**. Cloudless will work out of the box, but if we want Flipper to be communicating with Flipper cloud there are a couple of secrets that need to be set:
 
 - `FLIPPER_CLOUD_TOKEN` - needed for our App to be able to communicate with Flipper Cloud
 - `FLIPPER_CLOUD_SYNC_SECRET` - used if we want Flipper Cloud to send a webhook request to our app telling it to sync with Cloud
--- a/template.rb
+++ b/template.rb
@@ -1,3 +1,5 @@
+require 'tty-prompt'
+
 BASE_URL = 'https://raw.githubusercontent.com/infinum/default_rails_template/master'.freeze
 
 # Readme.md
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -35,6 +35,16 @@ For build workflow to work, the following secrets must exist (usually set up by
 
 For deploy workflows, you need to generate private/public SSH key pairs for each environment. Public key should be added to the server to which you're deploying. Private key should be added as a secret to GitHub and named `SSH_PRIVATE_KEY_#{ENVIRONMENT}`, where `ENVIRONMENT` is replaced with an appropriate environment name (`STAGING`, `PRODUCTION`, etc.).
 
+### Flipper
+
+This template uses Flipper for feature flag management. There are two ways for Flipper to function: **Cloud** and **Cloudless**. Cloudless will work out of the box, but if we want Flipper to be communicating with Flipper cloud there are a couple of secrets that need to be set:
```
uncoverd, line 40: Could you emphasize that the cloud version costs $$

template.rb
```diff
@@ -281,6 +281,38 @@
...
+        gem 'flipper-active_record'
+    HEREDOC
+  end
+
+  run 'rails g flipper:setup'
+end
+
+FLIPPER_CONFIG_FILE = <<-HEREDOC.strip_heredoc
+  Rails.application.configure do
+    config.flipper.memoize = false # for some reason webhook requests are being called twice without this
```
uncoverd, line 306: Did you find any related issue in the repository?

README.md
Benaaaaa: The cloud version is free though 🤔

template.rb
Benaaaaa: There is a PR that is talking about it, and I tried digging into it more, but nowhere a real solution seemed to be presented: https://github.com/flippercloud/flipper/pull/523

README.md
uncoverd: Oh my bad, I had a vague recollection of them only offering a paid cloud version 😓 Seems like it was introduced in August 2023, great 👍

uncoverd (approved):

README.md
vr4b4c, line 40: Cloudless sounds a bit out-of-place. Self-hosted seems to be a conventional name for this use case, but I realize authors also use that naming, so it's fine.

vr4b4c (changes requested):

README.md
nikajukic: Maybe add "self-hosted" in brackets next to cloudless so it's clear?

template.rb
nikajukic: Please add this link to the comment in the codebase.

template.rb
```diff
@@ -341,6 +373,9 @@
 
    redis_url: 'redis://localhost:6379'
 
+   FLIPPER_CLOUD_TOKEN: ''
+   FLIPPER_CLOUD_SYNC_SECRET: ''
```
nikajukic, line 377: This is secrets.yml, so shouldn't this be `flipper_cloud_token` and `flipper_cloud_sync_secret`, downcased as other parameters as well?

Benaaaaa: It seems a bit weird when reading it, but this is being added to the figaro file. Based on the documentation: 

> Using Environment Variables
Environment variables provide the easiest way to configure Flipper Cloud. Set a FLIPPER_CLOUD_TOKEN environment variable using [.dotenv](https://github.com/bkeepers/dotenv) or any other approach that works with environment variables, and Flipper will recognize it and handle the rest.
FLIPPER_CLOUD_TOKEN=your-cloud-token
FLIPPER_CLOUD_SYNC_SECRET=your-webhook-secret

By simply adding these two to the figaro file, figaro will load them as env variables upon startup. With such a setup, as long as the secrets are setup Flipper Cloud will work out of the box

template.rb
```diff
@@ -281,6 +281,38 @@
   end
 HEREDOC
 
+# Flipper
```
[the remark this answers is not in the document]
Benaaaaa: It seems that the flipper setup should not be run at this point. Will move it downards.

README.md
Benaaaaa: I was going with self-hosted first, but it seemed like if we only follow the setup process that it does not feel like we are hosting anything, but rather just using a cloudless approach. So the cloudless name seemed more appropriate for this. I will add the self-hosted next to cloudless for clarity purposes

template.rb
```diff
@@ -1,3 +1,5 @@
+require 'tty-prompt'
```
vr4b4c, line 1: No bueno, the template shouldn't depend on anything outside of the standard library. Currently, `rails new` fails if `tty-prompt` isn't installed beforehand
```
The template [https://raw.githubusercontent.com/infinum/default_rails_template/feature/960-introduce-a-feature-flag-management-system-to-the-default-rails-template/template.rb] could not be loaded. Error: cannot load such file -- tty-prompt
```

template.rb
vr4b4c: I don't think casing matters in practice, but I think we should keep it uniform. Since Figraro docs are using downcase and we followed it in other places, we should keep it consistent here as well, downcased.

template.rb
```diff
@@ -727,5 +763,10 @@ def run
 run 'overcommit --sign'
 run 'overcommit --sign pre-push'
 
+## Flipper setup for active record adapter
```
vr4b4c, line 766: I don't think we need this comment if the logic is self-explanatory.

vr4b4c (changes requested):

template.rb
Benaaaaa: I did introduce it downcased and it did not properly work downcased. Flipper cloud was not being loaded, it seems it is explicitly asking for upcased ENV variables

Benaaaaa: E.g. our figaro variables are set downcased: 
image
And we need them to be upcased for Flipper to work

template.rb
Benaaaaa: 😢

Benaaaaa:
@vr4b4c mostly just missing your review here, before this can be closed out

vr4b4c:
@Benaaaaa you have a conflict with the main branch

vr4b4c (approved):

PR #152, opened 2025-01-08
[Task #736] Add PR workflow to README

AnaCepic:
Task: [#736](https://app.productive.io/1-infinum/tasks/3497439)

#### Aim
Extract the NLBL workflow, combine with the survey results and add it to default rails template

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -39,108 +39,32 @@ bundle exec rspec
 
 ### Commits
 
-Commits should have a descriptive subject as well as a quick explanation of the reason for the change in the commit
-body.
-This makes it easier to check changes in the code editor as you do not have to find the pull request and open it
-on github.
-These commit bodies can also be used to fill the content of the pull request if you wish.
+Guidelines for writing commit messages are outlined in [this](https://infinum.com/handbook/rails/workflows/git/branches#other-important-notes-on-using-git) handbook chapter.
 
 ### Branches
 
-Branches should be opened from the master branch. Naming convention is {type}/{task-number}{descriptive-task-name}
-
-#### Types:
-
-- feature
-
-  A new feature or an improvement to an existing feature.
-
-- fix
-
-  A non-critical bugfix, improvement, paying the technical debt. Goes through code review process.
-
-- hotfix
-
-  A time sensitive critical bugfix, that should be deployed to production as soon as possible. Not necessary that it
-  goes through code review, but it should be revisited at a later stage, and properly fixed or improved.
+Our branch naming conventions are documented in [this](https://infinum.com/handbook/rails/workflows/git/branches) handbook chapter.
 
 ### Pull Requests
 
-Once the feature or fix is done, a PR for it should be opened. We have a pull request template with placeholders for all
-relevant data that should be included. Code-owners are automatically assigned as reviewers.
+See [this](https://infinum.com/handbook/rails/workflows/git/pull-requests) handbook chapter for pull requests guidelines.
 
 #### Labels
 
 We're using labels on PRs to visually mark the different states of the PRs. Some are self-explanatory, others have an
-additional description on github.
+additional description on GitHub.
 
 [TODO] Add list and descriptions of project-specific labels here.
 
 #### Solving Change Requests
 
-Change requests should be fixed in separate fixup or squash commits. Rebasing the branch during an ongoing review is not
-appreciated unless there is a good reason for it, like pulling in some new and necessary changes from master, because it
-makes harder for the reviewers to know what the new changes are and what they already reviewed.
+See [this](https://infinum.com/handbook/rails/workflows/git/clean-changes#solving-change-requests) handbook chapter for best practices on solving change requests.
 
-### Staging
+### Integration methodology
 
-Once the PR is opened, the feature can be merged into staging,
-unless it contains considerable logic in the migrations,
-in which case the reviewers should prioritize reviewing the migrations first
-and giving a thumbs up for a merge into staging.
+Check [this](https://infinum.com/handbook/rails/workflows/git/clean-changes#merging) handbook chapter for the integration methodology we use on the project.
 
-##### Merge Methodology
-
-We are doing merge-squashes to staging, as well of resets of the staging branch to master after each
-sprint, or more frequently as we see fit.
-
-We are usually doing merge squashes by cherry-picking a commit range.
-Cherry picking usually produces less merge conflicts once master and target branch diverge.
-
-```bash
-git switch staging
-git fetch origin staging && git pull --rebase
-git cherry-pick -n {BASE-OF-BRANCH}..{feature-branch}
-```
-
-Note that BASE-OF-BRANCH is one commit prior of the first commit of the branch.
-
-The commit message should be in the following format.
-
-```
-Merge-squash {feature-branch}
-
-(pull request {pull-request-link})
-[optionally](cherry picked from {commit-sha})
-```
-
-Including the pull request link makes github pick up that commit in the pull request, so we can know directly in
-the pull request, when was the branch deployed to staging.
-
-Run the specs, check if everything is ok, then push 🎉
-
-[TODO] If there is another environment present on the project, define the merge methodology for it.
-
-### Master (Production)
-
-Once the PR has at least 1 approval, the branch was successfully deployed to staging, tested,
-and there are no failing specs, it can be merged into master.
-
-##### Merge methodology
-
-Before merging we're also rebasing the feature branch on the latest master, so that the git history is nice and clean,
-and that the latest code can be run on CI before actually merging into master.
-We're also squashing the review comments corrections here, so make sure the `--autosquash` option is on by default
-or add that flag to the rebase command.
-
-[TODO]
-There are generally 2 ways we can merge PRs to Master:
-with `git merge --squash` (squash and merge button on GitHub) or non fast forward merge.
-Each of these is project specific and should be agreed upon which one would be used when starting the project.
-For more details on both metodologies, check [this](https://infinum.com/handbook/rails/workflows/git/clean-changes)
-Handbook chapter.
-Once you agree, you should replace this todo-paragraph with description of actual metodology being used.
-
+[TODO] Document any project-specific integration methodology if it differs from the one described in the handbook, or if the project includes additional environments.
 HEREDOC
 
 create_file 'README.md', README_MD, force: true
--- a/template.rb
+++ b/template.rb
@@ -35,6 +35,210 @@ Run:
 bundle exec rspec
 ```
 
+## PR Workflow
+
+### Commits
+
+Commits should have a descriptive subject as well as a quick explanation of the reason for the change in the commit
+body.
+This makes it easier to check changes in the code editor as you do not have to find the pull request and open it
+on github.
+These commit bodies can also be used to fill the content of the pull request if you wish.
+
+### Branches
+
+Branches should be opened from the master branch. Naming convention is {type}/{task-number}{descriptive-task-name}
+
+#### Types:
+
+- feature
+
+  A new feature or an improvement to an existing feature.
+
+- fix
+
+  A non-critical bugfix, improvement, paying the technical debt. Goes through code review process.
+
+- hotfix
+
+  A time sensitive critical bugfix, that should be deployed to production as soon as possible. Not necessary that it
+  goes through code review, but it should be revisited at a later stage, and properly fixed or improved.
+
+### Pull Requests
+
+Once the feature or fix is done, a PR for it should be opened. We have a pull request template with placeholders for all
+relevant data that should be included. Code-owners are automatically assigned as reviewers.
+
+#### Labels
+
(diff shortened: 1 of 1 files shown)


template.rb
```diff
@@ -35,6 +35,210 @@
...
+```bash
+{pr-title} ({pr-number})
+
+{pr-description}
+
+
+```
+
+*Note*: When doing merge-squash through terminal you will need to manually close the PR,
+as opposed to using the Squash and merge on GitHub
```
nikajukic, line 170: @uncoverd should we force to always use Github button for merging? Ideally we should have a rule in branch protection rules that no one should be able to push to master as everything should be done through PRs on Github. What do you think?

template.rb
```diff
@@ -35,6 +35,210 @@
...
+Once the PR has at least 1 approval, the branch was successfully deployed to staging, tested,
+and there are no failing specs, it can be merged into master.
+
+##### Merge methodology
+
+There are generally 2 ways we can merge PRs to Master:
+with `git merge --squash` (squash and merge button on GitHub) or non fast forward merge.
+Each of these is project specific and should be agreed upon which one would be used when starting the project.
+
+Once you agree you should update README accordingly.
```
nikajukic, line 131: I would somehow transform this into a [TODO] as ideally we'd keep only the information about the actual merge methodology in this Readme file. We can define both options in a Handbook, but only the relevant one should be kept in a readme.

vr4b4c:
Honest question, shouldn't this be a link to our Handbook chapter?

AnaCepic:
> Honest question, shouldn't this be a link to our Handbook chapter?

@vr4b4c I agree that most of the content is similar to what we already have in our Handbook chapter. I think it would be better to do what Nika mentioned: replace some parts with [TODO].

I’m not sure if you’re questioning the entire idea of adding this to the README 😅, but I do believe we should include this in our default template, even if we already have something similar defined in the Handbook.
From my experience, most developers tend to check the README file first when starting to work on a new project, rather than referring to the Handbook.
Having the workflow in the README also allows us to include project-specific details, such as environment-specific information, which can vary between projects.

That way we would have this information more accessible and it would also encourage people to clearly define these processes in the context of their project. This reduces the risk of important details being overlooked or vaguely considered.

AnaCepic:
> Honest question, shouldn't this be a link to our Handbook chapter?

@vr4b4c  I agree that most of the content is similar to what we already have in our Handbook chapter. I think it would be better to do what Nika mentioned: replace some parts with [TODO].

I’m not sure if you’re questioning the entire idea of adding this to the README 😅, but I do believe we should include this in our default template, even if we already have something similar defined in the Handbook. 
From my experience, most developers tend to check the README file first when starting to work on a new project, rather than referring to the Handbook.
Having the workflow in the README also allows us to include project-specific details, such as environment-specific information, which can vary between projects.

That way we would have this information more accessible and it would also encourage people to clearly define these processes in the context of their project. This reduces the risk of important details being overlooked or vaguely considered.

vr4b4c:
> I’m not sure if you’re questioning the entire idea of adding this to the README 😅,

Correct, I'm questioning the whole idea of having the sizeable amount of content about workflow repeated in projects README. A link(s) to the relevant chapter in the Handbook would be a better approach. But I'm open-minded towards new ideas, feel free to challenge my opinions.

> but I do believe we should include this in our default template, even if we already have something similar defined in the Handbook. From my experience, most developers tend to check the README file first when starting to work on a new project, rather than referring to the Handbook.

I'm not saying to omit the workflow documentation in the project repo, I'm saying we shouldn't repeat ourselves and link to some extent is more suitable for this use case (more on exact reasons follows). Let me give you some concrete examples of sections where I think link to the Handbook chapter is more suitable. 

### Commits
The commits section is very concise, generic (not project specific), and stable (not likely to change existing practices, albeit might be extended with some new ones). Keeping it centralized in the Handbook chapter would make it easier to manage and update. Since it's centralized, information propagation is free. I don't think it's likely for something project-specific to come up in this context, but if it does, the link would still be valid and the project would document only deviations from the centralized knowledge base.

### Branches
The same as for commits, handbook [chapter](https://infinum.com/handbook/rails/workflows/git/branches).

### Pull Requests
The same as for commits and branches, handbook [chapter](https://infinum.com/handbook/rails/workflows/git/pull-requests), which covers much more than the section in the default template.

### Solving Change Requests
This is best practice advice, I don't think it's likely to change on a per-project basis (handbook [chapter](https://infinum.com/handbook/rails/workflows/git/clean-changes#solving-change-requests)).

### Merge methodology
I do see this as something that's more likely to change than any other previously mentioned section. Still, I think it would be better to link it to Handbook [chapter](https://infinum.com/handbook/rails/workflows/git/clean-changes) and document only the differences to the established practice.

> Having the workflow in the README also allows us to include project-specific details, such as environment-specific information, which can vary between projects.

Linking generic topics shouldn't (and doesn't in my opinion) hinder our ability to document project-specific details.

> That way we would have this information more accessible and it would also encourage people to clearly define these processes in the context of their project. This reduces the risk of important details being overlooked or vaguely considered.

I don't think there is any difference regarding this concern if we link content or not. Depending on the change affiliation it should be added in the project repo (if it's project-specific) or changed in the Handbook chapter if it's globally applicable.

I'm now going to add a few comments (which should be fixed in the Handbook chapter), and think about how difficult would it be to propagate those changes if this PR was merged as is and applied to a dozen projects.

template.rb
```diff
@@ -35,6 +35,112 @@
...
 ```
 
+## PR Workflow
+
+### Commits
+
+Commits should have a descriptive subject as well as a quick explanation of the reason for the change in the commit
+body.
+This makes it easier to check changes in the code editor as you do not have to find the pull request and open it
+on github.
```
vr4b4c, line 45:
```suggestion
on GitHub.
```

template.rb
```diff
@@ -35,6 +35,112 @@
...
+These commit bodies can also be used to fill the content of the pull request if you wish.
+
+### Branches
+
+Branches should be opened from the master branch. Naming convention is {type}/{task-number}{descriptive-task-name}
```
vr4b4c, line 50:
```suggestion
Branches should be opened from the main branch. Naming convention is {type}/{task-number}{descriptive-task-name}
```

vr4b4c, line 50: Using folder-like structure in the branch names makes it more difficult to adopt git worktree's in practice. I'd personally suggest sticking to `{task-number}-{descriptive-task-name}` naming convention (this also matches naming strategy of task management systems that integrate with source control systems).

template.rb
```diff
@@ -35,6 +35,112 @@
...
+
+We're using labels on PRs to visually mark the different states of the PRs. Some are self-explanatory, others have an
+additional description on github.
+
+[TODO] Add list and descriptions of project-specific labels here.
+
+#### Solving Change Requests
+
+Change requests should be fixed in separate fixup or squash commits. Rebasing the branch during an ongoing review is not
+appreciated unless there is a good reason for it, like pulling in some new and necessary changes from master, because it
```
vr4b4c, line 82:
```suggestion
appreciated unless there is a good reason for it, like pulling in some new and necessary changes from the main branch, because it
```

template.rb
```diff
@@ -35,6 +35,112 @@
...
+makes harder for the reviewers to know what the new changes are and what they already reviewed.
+
+### Staging
+
+Once the PR is opened, the feature can be merged into staging,
```
vr4b4c, line 87:
```suggestion
Once the PR is opened, the feature can be integrated into staging,
```

template.rb
```diff
@@ -35,6 +35,112 @@
...
+unless it contains considerable logic in the migrations,
+in which case the reviewers should prioritize reviewing the migrations first
+and giving a thumbs up for a merge into staging.
```
vr4b4c, line 90:
```suggestion
and giving a thumbs up for integration into staging.
```

template.rb
```diff
@@ -35,6 +35,112 @@
...
+
+##### Merge Methodology
```
vr4b4c, line 92: The industry-adopted term for this procedure is "integration" and it covers any type of "joining" features, merging, rebasing, cherry-picking, ...
```suggestion
##### Integration Methodology
```

template.rb
```diff
@@ -35,6 +35,112 @@
...
+
+We are doing merge-squashes to staging, as well of resets of the staging branch to master after each
```
vr4b4c, line 94:
```suggestion
We are doing merge-squashes to staging, as well of resets of the staging branch to the main branch after each
```

template.rb
```diff
@@ -35,6 +35,112 @@
...
+The commit message should be in the following format.
+
+```
+Merge-squash {feature-branch}
+
+(pull request {pull-request-link})
+[optionally](cherry picked from {commit-sha})
+```
+
+Including the pull request link makes github pick up that commit in the pull request, so we can know directly in
```
vr4b4c, line 117:
```suggestion
Including the pull request link makes GitHub pick up that commit in the pull request, so we can know directly in
```

template.rb
```diff
@@ -35,6 +35,112 @@
...
+the pull request, when was the branch deployed to staging.
+
+Run the specs, check if everything is ok, then push 🎉
+
+[TODO] If there is another environment present on the project, define the merge methodology for it.
+
+### Master (Production)
```
vr4b4c, line 124:
```suggestion
### Main branch (Production)
```

template.rb
```diff
@@ -35,6 +35,112 @@
...
+
+Once the PR has at least 1 approval, the branch was successfully deployed to staging, tested,
+and there are no failing specs, it can be merged into master.
```
vr4b4c, line 127:
```suggestion
and there are no failing specs, it can be merged into the main branch.
```

template.rb
```diff
@@ -35,6 +35,112 @@
...
+
+##### Merge methodology
```
vr4b4c, line 129:
```suggestion
##### Integration methodology
```

vr4b4c (changes requested):

template.rb
Benaaaaa: Just to add here, if you add that rule - no pushing to master unless approved, if it is not satisfied, git won't let you push to master. On NLBL I use github, but some other people use the git CLI, and it works both ways for us.

AnaCepic:
@vr4b4c thanks for your comments! I've updated some parts to include a link to the handbook chapter and opened [PR](https://github.com/infinum/rails-handbook/pull/158) on the Rails Handbook to fix some of the issues you pointed out.

AnaCepic:
Also, @Benaaaaa, since the result of the poll you conducted was to use `cherry-pick -n` for staging and allow two approaches for production  (either `squash and merge` or `rebase and merge without squashing`), we should also update the Handbook chapter accordingly. 
Right now, it lists `merge-and-squash` or `cherry-pick` for staging, and `rebase-and-merge` (without squashing) for master. What do you think?

vr4b4c (approved):

vr4b4c:
Can you please port changes to https://github.com/infinum/rails-template as well 🙇 ?

template.rb
```diff
@@ -35,6 +35,36 @@
 bundle exec rspec
 ```
 
+## PR Workflow
+
+### Commits
+
+Guidelines for writing commit messages are outlined in [this](https://infinum.com/handbook/rails/workflows/git/branches#other-important-notes-on-using-git) handbook chapter.
```
cilim, line 42: absolute URLs means that when someone's reading a chapter on staging, they'll get redirected to the production Handbook when clicking the links :)

Please use relative URLs

template.rb
```diff
@@ -35,6 +35,36 @@
...
+
+### Branches
+
+Our branch naming conventions are documented in [this](https://infinum.com/handbook/rails/workflows/git/branches) handbook chapter.
```
cilim, line 46: ditto ^^

template.rb
```diff
@@ -35,6 +35,36 @@
...
+
+### Pull Requests
+
+See [this](https://infinum.com/handbook/rails/workflows/git/pull-requests) handbook chapter for pull requests guidelines.
```
cilim, line 50: ditto ^^

template.rb
```diff
@@ -35,6 +35,36 @@
...
+#### Labels
+
+We're using labels on PRs to visually mark the different states of the PRs. Some are self-explanatory, others have an
+additional description on GitHub.
+
+[TODO] Add list and descriptions of project-specific labels here.
+
+#### Solving Change Requests
+
+See [this](https://infinum.com/handbook/rails/workflows/git/clean-changes#solving-change-requests) handbook chapter for best practices on solving change requests.
```
cilim, line 61: ditto ^^

template.rb
```diff
@@ -35,6 +35,36 @@
...
+
+### Integration methodology
+
+Check [this](https://infinum.com/handbook/rails/workflows/git/clean-changes#merging) handbook chapter for the integration methodology we use on the project.
```
cilim, line 65: ditto ^^

cilim (approved):

minazek (approved):

PR #154, opened 2025-01-27
Add dependabot config for GH actions

lovro-bikic:
Adds dependabot configuration to open a PR when an action gets outdated (for example, when an [action gets deprecated](https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/)).

More info: [GH docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: 'github-actions'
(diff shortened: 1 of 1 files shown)


vr4b4c (approved):
🤘

PR #166, opened 2025-03-03
[Task #960] Introduce a feature flag management system

Benaaaaa:
Task: [#960](https://app.productive.io/1-infinum/tasks/7909978)

#### Aim
Flipper cloud free tier has changed its pricing - and as such is no longer compatible with out basic use case

#### Solution
Update template to generate setup for Flipper Cloudless

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,12 +37,25 @@ For deploy workflows, you need to generate private/public SSH key pairs for each
 
 ### Flipper
 
-This template uses Flipper for feature flag management. There are two ways for Flipper to function: **Cloud** and **Cloudless(Self-hosted)**. Cloudless will work out of the box, but if we want Flipper to be communicating with Flipper cloud there are a couple of secrets that need to be set:
+This template uses Flipper for feature flag management. There are two ways for Flipper to function: **Cloud** and **Cloudless(Self-hosted)**. This template sets up Flipper cloudless as free tier of Flipper Cloud does not cover the basic use case for our projects. 
+
+#### Flipper Cloudless setup
+
+Flipper Cloudless does not provide a dashboard and as such we are adding `flipper-ui` for the dashboard. In order to properly set up Flipper UI with basic authentication we need to set up two secrets:
+
+- `flipper_username_hash`
+- `flipper_password_hash`
+
+
+#### Flipper Cloud setup
+
+But, if we want Flipper to be communicating with Flipper cloud there are a couple of secrets that need to be set:
 
 - `FLIPPER_CLOUD_TOKEN` - needed for our App to be able to communicate with Flipper Cloud
 - `FLIPPER_CLOUD_SYNC_SECRET` - used if we want Flipper Cloud to send a webhook request to our app telling it to sync with Cloud
 
 Both of these secrets are available inside the Flipper Cloud dashboard, and need to be configured per environment
+We also would want to remove the `flipper-ui` gem, as Flipper Cloud provides a dashboard.
 
 
 #### Slack notifications
--- a/README.md
+++ b/README.md
@@ -43,8 +43,8 @@ This template uses Flipper for feature flag management. There are two ways for F
 
 Flipper Cloudless does not provide a dashboard and as such we are adding `flipper-ui` for the dashboard. In order to properly set up Flipper UI with basic authentication we need to set up two secrets:
 
-- `flipper_username_hash`
-- `flipper_password_hash`
+- `flipper_username`
+- `flipper_password`
 
 
 #### Flipper Cloud setup
--- a/template.rb
+++ b/template.rb
@@ -292,26 +292,16 @@ append_to_file 'Gemfile', after: /gem "rails".*\n/ do
     <<~HEREDOC.strip_heredoc
 
       gem 'flipper-active_record'
+      gem 'flipper-ui'
     HEREDOC
   else
     <<~HEREDOC.strip_heredoc
 
       gem 'flipper-redis'
+      gem 'flipper-ui'
     HEREDOC
   end
 end
-
-FLIPPER_CONFIG_FILE = <<-HEREDOC.strip_heredoc
-  Rails.application.configure do
-    config.flipper.memoize = false # for some reason webhook requests are being called twice without this, more on this link: https://github.com/flippercloud/flipper/pull/523
-
-    config.after_initialize do # if we are using webhooks, this will make sure our flags are synchronized upon initialization
-      Flipper.sync
-    end
-  end
-HEREDOC
-
-create_file 'config/flipper.rb', FLIPPER_CONFIG_FILE, force: true
 
 # Suppress Puma SignalException
 append_to_file 'config/puma.rb', after: /pidfile ENV.*\n/ do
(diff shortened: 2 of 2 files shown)


vlakre (approved):

Benaaaaa:
Ping @vr4b4c @nikajukic @uncoverd

vr4b4c:
Ciaos, sorry for the late reply, I'll take a look on Friday 28th

template.rb
```diff
@@ -751,5 +744,81 @@ def secrets
...
+    config.descriptions_source = -> (keys) do
+      {
+        # when creating a new feature flag, insert its name and description here e.g.
+        # 'new_feature' => 'This is a new feature'
+      }
+    end
+    config.show_feature_description_in_list = true
+  end
+
+  # If you want to Audit your Flipper enable/disable actions, uncomment the following code and create FlipperAuditLog table
```
vr4b4c, line 758: This looks half done to me. I'd ask user if he wants auditing and then add the whole package (config & migration).

template.rb
```diff
@@ -751,5 +744,81 @@ def secrets
...
+    builder.use Rack::Auth::Basic, "Flipper Admin" do |username, password|
+      next false if FLIPPER_USERNAME.blank? || FLIPPER_PASSWORD.blank?
+
+        ActiveSupport::SecurityUtils.secure_compare(Digest::SHA256.hexdigest(username), Digest::SHA256.hexdigest(FLIPPER_USERNAME)) &&
+        ActiveSupport::SecurityUtils.secure_compare(Digest::SHA256.hexdigest(password), Digest::SHA256.hexdigest(FLIPPER_PASSWORD))
+    end
+  end
+HEREDOC
+
+create_file 'config/initializers/flipper.rb', FLIPPER_CONFIG_FILE, force: true
```
vr4b4c, line 793: Should we skip `rails g flipper:setup` from this [line](https://github.com/infinum/default_rails_template/pull/166/files#diff-1ceb42e9dd4ccfd7b01826cb5e570004ec897791b4f43d0117d2a6e4d5717ea5R744) since it's going to be overwritten?

template.rb
```diff
@@ -751,5 +744,81 @@ def secrets
...
+  #       )
+  #     end
+  #   end
+  #
+  # end
+
+  #  ActiveSupport::Notifications.subscribe(/flipper/, FlipperSubscriber.new)
+
+
+  FLIPPER_USERNAME = Rails.application.secrets(:flipper_username)
```
vr4b4c, line 780: These are global constants, you might want to consider local variables (I'd even suggest moving them in the `builder.use` block).

template.rb
```diff
@@ -751,5 +744,81 @@ def secrets
...
+        ActiveSupport::SecurityUtils.secure_compare(Digest::SHA256.hexdigest(username), Digest::SHA256.hexdigest(FLIPPER_USERNAME)) &&
+        ActiveSupport::SecurityUtils.secure_compare(Digest::SHA256.hexdigest(password), Digest::SHA256.hexdigest(FLIPPER_PASSWORD))
+    end
+  end
+HEREDOC
+
+create_file 'config/initializers/flipper.rb', FLIPPER_CONFIG_FILE, force: true
+
+append_to_file 'config/routes.rb', after: 'Rails.application.routes.draw do' do
+  "mount Flipper::AuthenticatedApp, at: '/flipper'"
```
vr4b4c, line 796: Seems to me you are missing a newline. This way `config/routes.rb` ends up being
```ruby
Rails.application.routes.draw domount Flipper::AuthenticatedApp, at: '/flipper'
```

Replace string with multiline heredoc (you can find other cases of it being used in the `template.rb`)

template.rb
```diff
@@ -751,5 +744,81 @@ def secrets
...
+end
+
+append_to_file 'config/initializers/flipper.rb', after: 'ActiveSupport::Notifications.subscribe(/flipper/, FlipperSubscriber.new)' do
+  if flipper_storage_adapter == 'ActiveRecord'
+    <<~HEREDOC
+      # Flipper.configure do |config|
```
vr4b4c, line 802: How come this block (in the other conditional branch as well) is commented out?

vr4b4c (changes requested):

template.rb
Benaaaaa: The flipper:setup does not only setup the flipper.rb but also the migrations for active record adapter. That way we are ensuring we have our config while having the migrations that flipper:setup provides

template.rb
Benaaaaa: This is connected to your comment above 
> This looks half done to me. I'd ask user if he wants auditing and then add the whole package (config & migration).

This part is connected to the Auditing, but is dependent on the adapter we use. I see now that I am doing things in between (mounting the authenticated app route), which causes the confusion

template.rb
```diff
@@ -751,5 +744,104 @@ def secrets
...
+      Rails.application.configure do
+        ::ActiveRecord::Base.respond_to?(:retrieve_connection)
+      end
+
+    HEREDOC
+  end
+end
+
+if !no?('Do you want Flipper auditing? [Yes]', :green)
+  append_to_file 'config/initializers/flipper.rb' do
```
vr4b4c, line 786: What do you think about packaging this into a separate gem? It seems to be complicated enough to justify the cost of managing it separately. This would also make it appropriate for adding new features.

Currently, it's missing the model (`FlipperAuditLog`) definition used on line 789.

Additionally, can you please open PR for flipper support on https://github.com/infinum/rails-template ? Let me know if you have any questions regarding implementation.

Benaaaaa: Sure, will then for now open a new PR on rails-template that will include adding feature flag implementation. Open a task to create a gem, add it to personal goals, and schedule when I can work on creating a gem - in the meantime will be closing this PR, as we will be, I am assuming, moving to using rails-template

template.rb
```diff
@@ -751,5 +744,104 @@ def secrets
...
+            adapter = Flipper::Adapters::Redis.new(Redis.new(url: Rails.application.secrets.fetch(:redis_url)))
+            instrumented = Flipper::Adapters::Instrumented.new(adapter, instrumenter: ActiveSupport::Notifications)
+            Flipper.new(instrumented)
+          end
+        end
+      HEREDOC
+    end
+  end
+
+  run 'rails g migration CreateFlipperAuditLogs operation:string feature_name:string thing_value:string gate_name:string result:boolean'
```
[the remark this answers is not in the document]
Benaaaaa: I did not want the `rake db:migrate` to be executed immediately because It does not seem like we actually setup the database when this template is run - so db:migrate will break until we run bin/setup.

Once the user is ready to start using the app they will run db:migrate either way, or the app will tell them there are migrations that have not been run

README.md
```diff
@@ -37,12 +37,25 @@ For deploy workflows, you need to generate private/public SSH key pairs for each
...
+
+Flipper Cloudless does not provide a dashboard and as such we are adding `flipper-ui` for the dashboard. In order to properly set up Flipper UI with basic authentication we need to set up two secrets:
+
+- `flipper_username`
+- `flipper_password`
+
+
+#### Flipper Cloud setup
+
+But, if we want Flipper to be communicating with Flipper cloud there are a couple of secrets that need to be set:
```
minazek, line 52:
```suggestion
If we want Flipper to be communicating with Flipper cloud we need to set these secrets:
```

Benaaaaa:
Closing this PR as we are moving on using rails-template: https://github.com/infinum/rails-template/pull/15

PR #170, opened 2025-04-24
Add support for SSM to publish-docs workflow

vdasovic:
Publish docs workflow uses SSH to connect to the server - Added support for SSH-ing through SSM tunnel


lovro-bikic (approved):

vdasovic:
SSH config setup on runner as mina-dox does not support ssh_options like mina.

minazek (approved):

